PR #32, opened 2024-03-26
Fixed json with all sha-1 and fixed sonar in app build.gradle

CassioManuguerra:


Andy130604:
GGWP

Andy130604 (approved):
LGTM

CassioManuguerra:
ty bros it means a lot ^_^

zouyonghao:
Good patch! Also, try to fix the code format before merging.

CassioManuguerra:
> Good patch! Also, try to fix the code format before merging.

You are right this was a mistake due to the fact that we had not yet enforced a rule to prevent pushing without a passing CI and it will not happen again

PR #34, opened 2024-03-27
Bottomnavbar

ecornamu:
Implemented bottom navigation bar that allows switching between Map, Discover and Profile tabs with icons and texts ressources.
Probably doesn't follow the MVVM architecture faithfully yet but we can discuss that in person tomorrow


JoniDutoit (approved):
Great job ! The navigation bar looks good to me and it is really pratical that you're took a head start and created the 3 screens for profile, map and discover. Also, good job with the icons ! Maybe we can discuss about the discover icon, not sure if it matches perfectly the ideia of exploring/discovering, considering  a magnifying glass as icon could be an alternative. Otherwise, nicely done !

PR #36, opened 2024-03-28
Added functions to write/retrieve data on/from the Firestore DB (and now users are stored on DB when they login)

Thimphou:
1. **Overview of Changes**: Added Firebase Firestore integration to the `DatabaseFunctions` class for user data management.

2. **Functionality Added in DatabaseFunctions Class**:
   - `addUserToDatabase`: Registers user data in Firestore upon registration.
   - `getFieldFromUser`: Retrieves specific user data fields.
   - `addFieldToUser`: Adds new user data fields.
   - `updateFieldInUser`: Updates existing user data fields.
   - `createCollection`: Creates new Firestore collections.

3. **Users who login are now registered on DB**: Called my addUserToDatabase function where there is the authentication part.


npfrei (commented):
Read everything, no issues

Andy130604 (changes requested):
Should have the database package in model package to follow MVVM, otherwise everything is fine

Andy130604 (approved):
This is perfect, nice implementation :)

PR #37, opened 2024-03-29
Map/display map - Map screen and dataclass for points

JeremyHugentobler:
# PR: Add Map Feature and Data Class for Activity Display

## Overview
This PR introduces a new map feature to locate the activities easily in the application. Additionally, it includes the implementation of a data class responsible for representing activities on the map.

## API
I had to create an API key on google cloud in order to retrieve the tiles of the map and the key is for now stored in the project. 
The key has been restricted to this application in particular and linked to the SHA1 keys of each devs, so it is not necessary to find a way to store it as a secret (Even tough it is always better) 

## What's next
- Adding some UI testing / edge cases for the map 

- Internet connection issues -> what happens when no more wifi

- MVVM implementation: waiting for HILT to be set up in other part of the project

- User's position

### User position
I tried to implement it, but didn't find any good ressources on how to make it like I wanted. This part is essential for the app if we want to integrate some kind of itinerary, but the support with compose is not that complete.


npfrei (commented):
Seems to work fine, but will need to make API key not public

JoniDutoit (approved):
Great job ! The permissions seems great to me as the need for the phone's location and internet.  Otherwise, as pointed above, there are potential security problems (API key public) as detected by Sonar but I imagine this is only a prototype (to check if basic functionality works), so good job!

JeremyHugentobler:
Hi, I added a new commit to the PR  (#ead9ae0 )

## GPS position added to the map
When launching the map screen, a permission prompt is started and depending on your answer, the map shows a button on the top right corner to redirect you to your position. If you don't allow the location permission, the map stills work.
The default point is Lausanne, but would be nice to center on the user if he has accepted the localisation service.

## Refactoring
I added some state to the viewmodel to make the map more dynamic, which will be helpful when needing to place more markers.

## TODO:
- Find a way to delay the first load of the map to directly show the user position if he accepts.
- Move the permission code in the VM (outside of a composable, but difficult)
- Hide the API key elsewhere even if it is restricted and not that big a deal right now
- Center the camera on the user

CassioManuguerra:
> Hi, I added a new commit to the PR (#ead9ae0 )
> 
> ## GPS position added to the map
> When launching the map screen, a permission prompt is started and depending on your answer, the map shows a button on the top right corner to redirect you to your position. If you don't allow the location permission, the map stills work. The default point is Lausanne, but would be nice to center on the user if he has accepted the localisation service.
> 
> ## Refactoring
> I added some state to the viewmodel to make the map more dynamic, which will be helpful when needing to place more markers.
> 
> ## TODO:
> * Find a way to delay the first load of the map to directly show the user position if he accepts.
> * Move the permission code in the VM (outside of a composable, but difficult)
> * Hide the API key elsewhere even if it is restricted and not that big a deal right now
> * Center the camera on the user

Great job on these functionalities, which seem very relevant to me. Hiding the API key elsewhere seems also pretty important to handle

JeremyHugentobler:
I found a way to hide the API key, so to display the map you can:

- Ask me to securely give you the API key
- Go find it in the google cloud admin panel 

Once you have the key, go to the local.properties file in your project and add a line in the form:
MAPS_API_KEY=NIOFIJNDNISFNIDSNF (which is not a real key in this example ^^)

CassioManuguerra:
# Title

PR #38, opened 2024-03-29
User/Profile Page - User's informations and activities are displayed and interactive

JoniDutoit:
# PR: Displaying basic user's informations interactively

## Overview
This PR includes a new starting look for the User's Profile screen as well as a helper class BarGraph.tk for displaying graphs with JetPack Compose. 

## Profile Screen

Displaying basic user's informations as photo, name and bio as well as a prototype chart for user's tracking their progress overtime

## Helper Class BarGraph

**Functionality added by this class**:
   - `BarGraph`: A composable function that draws a bar graph.
   - `BarType`: Enum class for the type of bar in the graph.
   - `WeekDay`: Enum class for the days of the week.

## What's next
- Implementing the rest of the components of the user's profile page following the figma sketch

- Internet connection issues -> what happens when no more wifi

- MVVM implementation: waiting for HILT to be set up in other part of the project

- Testing for coverage code

- Making the page more interactive to changes alongside with the database


Thimphou:
Looks good to me! Good job Joni

JeremyHugentobler:
Hi, 
Nice work ! Here are some suggestions for your PR and your "What's next" paragraph:

- After clicking on the "add a few word about yourself" textbox, it would be better if the placeholder is cleared so that we can directly write in it and not having to delete the previous content.
- For the "What happens when no internet", the sign in part of the app will most probably cache the information about the user on local storage, which could be used for this page. 

It is a good start and the work you have done on the figma sketch is really good !!

npfrei (commented):
Great, still needs to be implemented in NavController

CassioManuguerra (commented):
Good job ! 
The code looks good, well implemented, on a more conceptual aspect, maybe some other parameters can be added later, such as 
- setting the user's level in the parameters so that a targeted search can be made and the user can avoid too difficult activities for example
- allowing for a small description about the user
- an option to put the user's email public so that other users can contact him/her and maybe join for activities

app/src/main/java/com/lastaoutdoor/lasta/data/model/UserModel.kt
```diff
@@ -0,0 +1,8 @@
+package com.lastaoutdoor.lasta.data.model
+
+data class UserModel(
```
Andy130604, line 3: We still have to discuss this class, this is real hard to make right

settings.gradle.kts
```diff
@@ -19,6 +19,6 @@ dependencyResolutionManagement {
     }
 }
 
-rootProject.name = "Lasta"
+rootProject.name = "lasta"
```
Andy130604, line 22: Is this intentional, does this affect anything else?

Andy130604 (approved):
Overall really nice work @JoniDutoit, very appreciative of all the different types you created to fit your original conception. Also I don't understand the change in _settings.gradle.kts_ file.

settings.gradle.kts
JoniDutoit: No! It was a mistake, actually struggled with linking the project name to the app in my computer and I had to change it to make it work, now it's fixed !

PR #39, opened 2024-03-29
API calls and classes used to represent activities

npfrei:
Handles API calls from Overpass Turbo (OSM data)
Added Files : 
ApiResponse.kt : used to handle asychronous API calls
ApiService.kt : interface used to formulate specific API calls
OutdoorActivityResponse.kt : class used by gson to convert api responses to Kotlin classes
ActivityType.kt : enum class to reprensent the type of different activities (e.g. climbing, hiking, .. )
OutdoorActivity.kt : reprensents most general instance of an activity
Node.kt : represents an OSM node, inherhits from OutdoorActivity
OutdoorActivityRepository.kt  : Class used to get OutdoorActivities from overpass API (i.e. class used to make requests)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -157,6 +157,10 @@ dependencies {
 
     // ----------       Robolectric     ------------
     testImplementation(libs.robolectric)
+    // ----------     Retrofit     ------------
+    implementation("com.squareup.retrofit2:retrofit:2.9.0")
+    implementation("com.google.code.gson:gson:2.10")
+    implementation("com.squareup.retrofit2:converter-gson:2.9.0")
 
     //navigation with compose
     val nav_version = "2.7.7"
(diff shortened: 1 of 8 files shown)


Thimphou:
Great job! Maybe next step is to get and store the activities from/in the Firestore DB?

JeremyHugentobler:
You did a good job implementing the API calls, it seems to be working well.
I think the next step would be to start creating some methods to fetch activity with certain constraints like
- getActivities(POSITION, RADIUS, LIST<ActivityType>)

A small page in the wiki with some explanation on how to use the functions could be also a good feature, but otherwise good job thank you !

CassioManuguerra (approved):
Looks great ! Now we have to synchronize this with the app display because I have some files that overlap with yours such as  
- OutdoorActivity.kt
- OutdoorActivityRepository.kt

which I pre filled trying to anticipate what you would get from the API

PR #42, opened 2024-04-03
Added a Discovery Screen

CassioManuguerra:
Implemented the discovery screen as well as a structure around it, by creating model.data.OutdoorActivity.kt, model.repository.OutdoorActivityRepository.kt, view.screen.DiscoveryScreen.kt, viewmodel.OutdoorActivityViewModel.kt


app/src/main/java/com/lastaoutdoor/lasta/model/data/OutdoorActivity.kt
Andy130604, line 1: Should add nullable type safety, in case some activity is either not retrieved or has missing data.

Andy130604 (approved):
This is some really nice work, I like the use of the repository design pattern, improves overall code understanding. Still have to implement Hilt to link everything with the rest of the app but otherwise everything LGTM :)

app/src/main/java/com/lastaoutdoor/lasta/model/data/OutdoorActivity.kt
CassioManuguerra: Good comment, will keep that in mind !

PR #66, opened 2024-04-08
Map API Call for Climbing + Fix of the CI

JeremyHugentobler:
## Content of the PR

The CI script was blocked due to the lack of testing files, we are currently working on creating real tests, but in order to have a fixed main branch, I added some dummy tests files.
I also implemented calls to the API in the viewmodel to fetch some climbing activities on the map.

## To continue
- Find a way to make non-blocking calls to the API (now the map is a bit laggy)
- Find a way to properly display itinerary -> way to much node to display them on the map directly


Andy130604 (approved):
This is a fondamental PR for our project, nice work Jeremy LGTM

PR #69, opened 2024-04-08
Discovery/moreinfo

ecornamu:
### Implemented Pop-up Dialog for Activity Information

-Implemented a pop-up dialog that displays information on an activity when the user clicks its "more info" button. At the moment, it only displays the attributes of the `OutdoorActivity` class, but it can be linked to the activities database later on.


CassioManuguerra (approved):
Nice implementation, could have made this composable on a new page instead of using visibility, but functions well. This will have to be modified soon though, as the different features we display for an outdoor activity will be revised. Good job !

JeremyHugentobler (changes requested):
Hello,

Thank you for the dialog box, the composable works on my emulator and displays the correct informations, but regarding the structure, I'm requesting 2 changes:
- Use the view model of the discovery page to store the state you are using in your PR. All logic should be as much as possible written in View Model files.
- Add more comments to your code. Composable are hard to comment but adding lines describing what your function will display and what your states represent would be helpful

ecornamu:
> Hello,
> 
> Thank you for the dialog box, the composable works on my emulator and displays the correct informations, but regarding the structure, I'm requesting 2 changes:
> 
> * Use the view model of the discovery page to store the state you are using in your PR. All logic should be as much as possible written in View Model files.
> * Add more comments to your code. Composable are hard to comment but adding lines describing what your function will display and what your states represent would be helpful

Thank you both for your feedback !
As asked by @JeremyHugentobler , here are the changes I made : 
- The code logic now resides in the `DiscoveryScreenViewModel` so that there is only `@Composable` code in `DiscoveryScreen`
- The code is now commented for better readability

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/DiscoveryScreenViewModel.kt
```diff
@@ -46,4 +49,25 @@ constructor(private val repository: OutdoorActivityRepository) : ViewModel() {
               node.tags.name ?: "Unnamed Climbing Activity"))
     }
   }
+  /*Toggles the dialog box on/off*/
+  val displayDialog = mutableStateOf(false)
+  /*Just a default activity to fill in the mutable state*/
+  val dummyActivity = OutdoorActivity(ActivityType.HIKING, 3, 5.0f, "2 hours", "Zurich")
+  val activityToDisplay = mutableStateOf(dummyActivity)
+
+  @Composable
```
Andy130604, line 58: Should not have composable in viewmodel. This is bad practice

Andy130604 (changes requested):
Look at comments changes should be implemented for app consistency

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/DiscoveryScreenViewModel.kt
ecornamu: Done

JeremyHugentobler (approved):
Thank you for your changes, 
LGTM now !

Andy130604 (approved):
Nice changes, nice feature implementation. You can merge now

PR #74, opened 2024-04-09
Map: Info when clicking on a marker on the map

JeremyHugentobler:
## Content
Added what is called a Modal Bottom Sheet that appears when you click on a specific marker of the map. The data fetched from the OSM API call is written to this sheet.

Also did some general refactoring of the MapScreen -> created smaller functions

## What's next
- Not related to the PR but hiking activities and itineraries are not easily drawable, we need to rethink how the data is fetched and how to display it efficiently
- Get better data to display, for now there is only an activity name and type -> implement the DB on our side and guarantee better information sources
- Keep refactoring the view, still not ideal but better

### DIFF
--- a/app/src/main/java/com/lastaoutdoor/lasta/ui/screen/MapScreen.kt
+++ b/app/src/main/java/com/lastaoutdoor/lasta/ui/screen/MapScreen.kt
@@ -1,12 +1,25 @@
+@file:OptIn(ExperimentalMaterial3Api::class)
+
 package com.lastaoutdoor.lasta.ui.screen
 
 import android.Manifest
 import android.annotation.SuppressLint
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.contract.ActivityResultContracts
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.ModalBottomSheetValue
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.ModalBottomSheet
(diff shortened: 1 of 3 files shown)


ecornamu:
- Clever way of displaying the data of the activities on the screen without hiding the map
- Your code is incredibly modular and well commented, it is really perfect to read (I need to conform to this example for mine haha)

_Regarding your later comments_ : As discussed IRL, I agree that we need to make a decision on the database situation which is really a pain to use at the moment

ecornamu (approved):
-Clever way of displaying the data of the activities on the screen without hiding the map
-Your code is incredibly modular and well commented, it is really perfect to read (I need to conform to this example for mine haha)

_Regarding your later comments_ : As discussed IRL, I agree that we need to make a decision on the database situation which is really a pain to use at the moment

CassioManuguerra (approved):
Hi, 
This is a nice feature, it looks better than the previous small text boxes that were appearing. It is working nicely, but is relying on an old MapViewModel which is not using hilt, so keep in mind that we will probably have to come back to this code and adapt it for the next steps. The comments are very clear and present, good job on this!

CassioManuguerra:
As for what's next, we will indeed probably have to rethink how we get our activities because OSM is clearly not ideal, we are lacking important information.

Andy130604 (approved):
Nice commenting of your functions, I would suggest to decompose your composable files into smaller items that are easier to work with. Otherwise everything LGTM

npfrei (approved):
Great, the comments are very clear and the code seems fine. I would suggest you add things like distance to the user's position, as it is quite simple and doesn't depend on the API or the DB

PR #75, opened 2024-04-09
Functions to retrieve/write data on activities_database on Firestore.

Thimphou:
### DatabaseManager class

I changed the class "DatabaseFunctions" with a more generic name : DatabaseManager. 

This class has now two new functions to manage on activities_database collection on Firebase. 

- `getFieldOfHiking`: Retrieves a field of a hiking activity (such as averageSpeedInKMH for example). Keep in mind that you need to specifiate which field you want to get as a String parameter. 
- `setFieldOfHiking`: Sets a field of a hiking activity. Same as above for the usage. 
- `getUserPreferences`: Gets the UserPreferences refered to this user. 
- `updateUserPreferences`: Updates a user's preferences in the Firestore database. 

### UserModel class

- Added constructor to create UserModel from the FirebaseUser, to facilitate proper adherence to the UserModel that has been created previously. 
- Added UserPreferences as well, and adapted it in GoogleAuth.


ecornamu (changes requested):
Good job ! This batch of code seems to implement all the basic functions that we will need to manipulate the database as we want and use those functions to build more specific ones. 

I have 3 remarks/questions on your code : 
1. You seem to have also pushed my changes regarding the `More Info` button that come from another branch. Is it normal ? Accepting your PR would then mean accepting mine (that is still ongoing) at the same time
2. Since every field is refered to as a `String` value, could it be clever to add an `enum` with possible field names so that an error for a wrong field name cannot happen ?
3. In my opinion the code could be a bit more commented to understand the database calls ;)

I would accept your PR but I'll instead request changes so that we tackle this issue of my changes from another PR being in yours

Thimphou:
The "enum" idea sounds smart, I'll definitely implement it tomorrow! 

> Good job ! This batch of code seems to implement all the basic functions that we will need to manipulate the database as we want and use those functions to build more specific ones.
> 
> I have 3 remarks/questions on your code :
> 
> 1. You seem to have also pushed my changes regarding the `More Info` button that come from another branch. Is it normal ? Accepting your PR would then mean accepting mine (that is still ongoing) at the same time
> 2. Since every field is refered to as a `String` value, could it be clever to add an `enum` with possible field names so that an error for a wrong field name cannot happen ?
> 3. In my opinion the code could be a bit more commented to understand the database calls ;)
> 
> I would accept your PR but I'll instead request changes so that we tackle this issue of my changes from another PR being in yours

CassioManuguerra (commented):
Hello,
Your implementation will nicely suit the direction that our app will take, well done !
A few comments :
- We will probably need functions to set the difficulty level of a hike
- We are thinking of filling our own database of activities, and we will probably have to add some functions to store information about an activity.
- Optionally, it would be a nice touch to be able to store comments about a trip, keep this idea in case we have time

Otherwise good commenting, and the functions are good !

CassioManuguerra (approved):

ecornamu (approved):
- The code is now well commented and thus more readable
- You added the mentionned enum to avoid faulty string inputs
- You removed the changes coming from another PR

Life is good and so are you :)

PR #79, opened 2024-04-10
Fix: Marker were not fetched when moving the camera

JeremyHugentobler:
### Bug description
After a big restructuration of the project, a functionnality of the map was affected: markers were not displayed anymore when the camera was moved around.

### Fix
It was an extremly dumb fix, that took me more than 2 hours to find... During the last merge, someone must have removed an elvis operator from the MapViewModel that was checking whether the activities had a valid name ( not null ) and replaced it with a placeholder otherwhise. ( The API data structures were not allowing null fields, which raised warnings about the Elvis operators. I added the '?' after the concerned fields, which resolved the warnings)

This error never crashed the program, because it was in a try catch destined for the API calls. The error was only printed in the console which made it hard to find and I was convinced it was coming from a more obscure part of the code.

### Next
I spent a considerable amount of time looking for solutions and realized that our implementation could be a lot better (new repository only for markers, different state for the asynchronous request to the api (@npfrei we can look into it together) ) and I will definitively try to ameliorate it in the future.

Moreover error handling is something we should start to implement seriously. For example show an in app message when there are some networking issues

### DIFF
--- a/app/src/main/java/com/lastaoutdoor/lasta/data/model/activity/OutdoorActivity.kt
+++ b/app/src/main/java/com/lastaoutdoor/lasta/data/model/activity/OutdoorActivity.kt
@@ -11,10 +11,10 @@ open class OutdoorActivity(
     val length: Float,
 
     /** The estimated duration of the activity. */
-    val duration: String,
+    val duration: String?,
 
     /** The canton and postal code of the activity. */
-    val locationName: String,
+    val locationName: String?,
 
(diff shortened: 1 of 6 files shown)


Andy130604 (approved):
The error about Elvis operator is really unfortunate. I fully agree with you that error handling is a big issue right now and we will have to look into it in the future. Otherwise the view models are starting to look good. Very nice editing 👌

PR #82, opened 2024-04-11
Implemented testing in authviewmodel and started on authrepository

Andy130604:


CassioManuguerra (approved):
Hello,
these tests look functional, good work ! It will not be contributing a lot to the coverage since these are small unit tests, but we're on track :-)
For next PR maybe fix the tests that you commented for now it's fine but we will need every percent we can get ^_^

PR #83, opened 2024-04-11
Testing/api

npfrei:
Added Tests for Classes related to API calls, gives 100% line coverage


app/src/test/java/com/lastaoutdoor/lasta/data/api/ApiDataClassesTest.kt
```diff
@@ -0,0 +1,215 @@
...
+import com.lastaoutdoor.lasta.data.model.map.Position
+import com.lastaoutdoor.lasta.di.AppModule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class ApiDataClassesTest {
+  private val con = ApplicationProvider.getApplicationContext<android.content.Context>()
+  private val rep = OutdoorActivityRepositoryImpl(AppModule.provideAPIService(con))
```
ecornamu, line 21: Maybe it would be more optimal to set those values as `private lateinit` and then assign their values in a `@Before` set up function, but that's not urgent

ecornamu (approved):
Well done ! This seems like a complete code coverage for the API data class. Not much to add except for the comment on test variables :)

PR #84, opened 2024-04-11
Tested the map ViewModel

JeremyHugentobler:
## Content
I wrote some tests for the Map ViewModel by mocking a fake repository to simulate the activity API calls. 96% Coverage obtained.

## The remaining 4%
I sadly could not achieve 100% due to mutableStateOf(MapState()), who was not analysed correctly by Jacoco (when modified, it wasn't recorded as a call to the setState function in the kotlin object). I will look into it later, but it is a small portion of code and not at all critical for our app, since this value should not be modified to much


CassioManuguerra (changes requested):
Check your CI, it is not passing #KTFMT

PR #87, opened 2024-04-11
Testing/profilestats

ecornamu:
Added testing to two folders, mainly : 
- Fully tested data/model/profile folder which amounts to 85% line coverage. The remaining 15% is trivial code such as `enums` and `data classes` with no additional functions. I can add test for them if we really seek full coverage but I don't think it is relevant
- Tested  the `Activity Converter` for a line coverage of 99%, which amounts to 28% of the lines of the data/db folder


npfrei (approved):
Great, all tests pass

PR #97, opened 2024-04-12
First android test for the map... maybe the last

JeremyHugentobler:
## Content
I added some tests for the bottomsheet, but this is basically all I was managed to test on the map composable
## Issues
The Google Map Composable and other part of this screen depends are not easily mockable (android permission pop-up, the map in itself, ...) and this block the coverage of this part. The map needs an API key, that is not present in any form in the CI, which will then not be tested. 
For the permission pop-up, I didn't find any documentation online on how to test such kind of components.


ecornamu (approved):
The tests all pass :) thank you

PR #98, opened 2024-04-12
Wrote a workflow file to build apk and store it in Artifacts

Andy130604:
I added the Maps API in github secrets since I we need it for building the apk.


ecornamu (approved):
Read and approved by the council

PR #108, opened 2024-04-12
Bug fix in the profile user screen

JoniDutoit:
# PR: Fixed a bug that in the profile screen

## Overview
Fixed the bug where when you tried to change the sport type at the same time as fetching recent activities actually crashed the application.


ecornamu (approved):

Thimphou (approved):
Nice bug fix

PR #118, opened 2024-04-16
Made a working apk building workflow script and added some security by removing google json file from project

Andy130604:


CassioManuguerra (changes requested):
Hi, 
your PR seems to have affected the whole project as if it had been rewritten, not sure why it happened but it would be nice to know in order to avoid this in the future, as it can lead to some bugs getting infiltrated, because as reviewers we aren't going to go through the 106 files changed to make sure no line has been added somewhere by mistake.

The code you added is good, very nice to be able to finally build the apk correctly ! There is something that you could do a bit more though, and that is to comment a bit more your code, even if it seems straight-forward in `.github/workflows/apk-build.yml`and `.github/workflows/ci.yml`, at least in `app/build.gradle.kts`  ! (and even for the other two, commenting very obvious code is never bad)

Apart from that, good job !

ecornamu (approved):
Good job ! The changed files are now correctly displayed by github and only revolve around the apk workflow implementation. LGTm

CassioManuguerra (approved):
Thank you for your changes, the comments are clean, we can really feel that you poured your soul into them. Nice job :-)

PR #126, opened 2024-04-18
User SignIn now consistent with DB and SetupScreen implemented. 

Thimphou:
### Description
This PR proposes changes to the `SetupScreen` composable function in the `com.lastaoutdoor.lasta.ui.screen.setup` package. Here is what it looks like! 

Now you arrive on this page when you log in (Note: in the future, it will only be at the first connection of the user to Lasta, so when the user signs up). You can select your prefered language and sport (for the moment between Hiking and Climbing). 
Then, when you click on "Save", it will save these parameters in the user's document in Firestore, and redirect you to the main screen :)

## AuthRepositoryImpl

There is now a difference between a SignUp (first time on the app) where you write user's data in the DB, and a SignIn where you read the already existing data. 

## AppModule

Modifications to achieve the same goal as above, i.e distinguishing between Sign In and Sign Up. 

## DatabaseManager 

Great return of our beloved DatabaseManager class, which will soon be separated in an interface and some implementations to be more precise, instead of having everything in the same class. 

## UserModel

Note that now the UserModel only stores the hikingLevel instead of the all UserPreferences object. (Note that I adapted the unit tests of PreferencesViewModel class). 

## LoginNavGraph

Created this class to navigate easily between all the login screens (i.e LoginScreen and SetupScreen). 

## MainAppNavGraph

Now the MainAppNavGraph takes the LoginNavGraph as the Login Route to access both LoginScreen and SetupScreen easily, instead of only LoginScreen before! 

## Routes

There are now new routes to access those two screens mentionned above. 

## LoginScreen

Uses the now Leaf Login route instead of a root one, due to the modifications described above. 

## ProfileScreen

Parameters on the profile screen now are stored on the DB. For example, if you change the "0" value to "1" (so from HikingLevel.BEGINNER to HikingLevel.INTERMEDIATE) as shown below, it is also stored on your user document in the database.
 

## Conclusion

Hope this changes are improving the app!


CassioManuguerra (changes requested):
Hi Maxime, or should i say "ayo wassup dawg" ? 
Your PR features solid changes ! Here's a brief review and some suggestions for next steps:

- Setup Screen: The implementation of language and sport selection, along with data saving functionality is a great step forward. Nice job !
- AuthRepositoryImpl: Distinguishing between SignUp and SignIn operations is important for effective user data management. As a next step, it would maybe be smart to add more error handling for various authentication scenarios to provide better feedback to the user and avoid crashes.
- Routes: Introducing new routes for improved navigation flexibility is great ! We could imagine implementing route guards for specific access control later.

There is a notable problem though, which is that you let some `println` in the final code, which we can't push on the main. Please fix this before we approve the pull request.
But apart from that nice job !

CassioManuguerra (approved):
Thank you for editing your code, the `println` are gone and I can sleep on both my ears :)

PR #135, opened 2024-04-19
Discovery screen refactor to include the map and match figma

JoniDutoit:
# PR: Adapted the discovery screen to include the map inside and also nicer UI design as the Figma Mockup

## Overview
Added a new package for components that can be reused all over the project. Re-implemented the discover screen to include the Map screen as well as UI changes according to the Figma design

## Discover Screen
Has a lot of new functionalities (still has many buttons to be implemented in order to be functional), new design and includes the map screen

## Discover Screen ViewModel
Adapted because the dialog of the activity selected will be replaced for a whole screen. Furthermore, added a state to keep track of wheter the list of activities or the map is displayed adapting the UI accordingly. 

## Package components
New package that contains all common/reused components through out the project
- DisplaySelectorComponent: display an horizontal navigation, used to select the screen on the discover screen or the time frame in the profile screen
- DropDownMenuComponent: display a drop down menu when clicked on with a text parameter as title
- SearchBarComponent: new component!! Search bar used in the discover screen profile for searching specific activities
- SeparatorComponent: an horizontal separator for creating layers in the UI.

## What's next

- Test new implemented funcionalities

- Start implementing buttons functionalities

- Complete text fields for activities as popularity, distance from location and difficulty.

https://github.com/LASTA-OUTDOOR/lasta/assets/95919886/7107a8d7-5afe-4ff6-8e5f-39f3b8d8f853


Andy130604 (approved):
This is insane work, you really have a sense for UI dev. We will have to test the profile screen a bit more for coverage.

PR #160, opened 2024-04-20
Map - Display itinerary only when clicked

JeremyHugentobler:
## Features
I changed some features of the map to make it more readable
- When clicking on a marker, the camera is centered on it
- If the clicked marker is linked to an itinerary, it'll be displayed on the map after the click
- Good coverage on the viewmodel, but UI is still almost impossible to test...

## Next
My changes make the content visually cleaner, but since we still haven't fixed our API calls issues, the data is still completely requested, even if not showed. (Performance issues)

Some part of the Screen should be refactored (the GoogleMap composable is quite big), but every time I try something to modularize it, it creates an issue with recomposition (Display marker when loaded, but the trigger when moving the camera doesn't work)


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/map/MapScreen.kt
Andy130604, line 1: Nice implementation of the itinerary display. You also fixed the fact that the markers and weren't loading on screen compose without moving the map.

Andy130604 (approved):
Overall, really well documented code. Could improve on **readability** in composables sometimes and you also added yourself a todo to take out the added markers and itineraries from `MapScreen.kt` file to make more modulable for future use.

PR #163, opened 2024-04-21
Implementing basic search options to target activities

CassioManuguerra:
# PR

## Overview
This PR features search options allowing the user to direct his/her activity searching, both for list mode and map mode. 

https://github.com/LASTA-OUTDOOR/lasta/assets/91049597/6960a5c2-8378-4b70-907a-8bf3397e1804 

map search options
map search options light

## Changes
Added a popping menu triggered by clicking on the down-faced arrow next to the Locality displaying, allowing to choose from different localities, as well as choosing the range inside which the activities will be shown. By clicking on the "Apply" button, the changes are applied and a subsequent call to the API (and when finished to the database too) is made, resulting in a display of the activities found. The new saved search settings are then displayed on the top left of the screen. 
### Files changed
- `app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discovery/DiscoveryScreen.kt` : Added components that display the search options and call the functions from `DiscoveryScreenViewModel` with the updated parameters.
- `app/src/main/java/com/lastaoutdoor/lasta/viewmodel/DiscoveryScreenViewModel.kt` : Added variables and setters to store range and position settings.
- `app/src/androidTest/java/com/lastaoutdoor/lasta/ui/screen/discovery/DiscoveryScreenTest.kt` : Added android tests to check for the components correct functioning.

## What's next
Next steps would be to 
- Implement more options in the menu
- Finish implementing the visual searching on the map mode
- Link with database when implemented

## Notes
The search options for map mode are not yet linked with the map view model because it is currently under changes and will be working differently.

### DIFF
--- a/app/src/androidTest/java/com/lastaoutdoor/lasta/ui/screen/discovery/DiscoveryScreenTest.kt
+++ b/app/src/androidTest/java/com/lastaoutdoor/lasta/ui/screen/discovery/DiscoveryScreenTest.kt
@@ -105,6 +105,16 @@ class DiscoveryScreenTest {
     composeRule.onNodeWithTag("localitySelectionDropdownButton").performClick()
   }
 
+  // Test the modal upper sheet
+  @Test
+  fun modalUpperSheet_isDisplayed() {
+    composeRule.activity.setContent {
+      DiscoveryScreen(navController = rememberNavController(), discoveryScreenViewModel)
+    }
+    composeRule.onNodeWithTag("listSearchOptionsEnableButton").performClick()
+    composeRule.onNodeWithTag("modalUpperSheet").assertIsDisplayed()
+  }
+
   /*
     // Test if discovery content is displayed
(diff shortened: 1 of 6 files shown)


Andy130604 (approved):
This is some amazing work @CassioManuguerra, you've split up your code into different composables and documented and tested them and the result looks Great! 👌 The one thing I could think off is maybe divide the `DiscoveryScreen.kt` composables into 1 or 2 different files into a `components` folder inside `discovery` package.

CassioManuguerra:
> This is some amazing work @CassioManuguerra, you've split up your code into different composables and documented and tested them and the result looks Great! 👌 The one thing I could think off is maybe divide the `DiscoveryScreen.kt` composables into 1 or 2 different files into a `components` folder inside `discovery` package.

Thank you for your feedback Andrew ! You're right it would be better separate my components in another file I will do that now.

PR #164, opened 2024-04-21
Update API calls and also make them asynchronous

Andy130604:
## Description:
This PR updates the current API calls to overpass since they were synchronous and were blocking UI composition while running. Also, this PR aims to optimise overall app's data networking by making the calls with reduced data when only fetching the activities metadata (not fetching the geographical data results in much smaller request responses size).

## Task List:

**Code Implementation**
  - [x] Create all OSM API queries that app needs
  - [x] Implement Queries with RetroFit
  - [x] Write asynchronous functions to call the queries depending on needed data
  - [x] Connect everything with the current viewmodels in place to make sure the new version doesn't break everything

 **Documentation**
  - [x] Document the queries to explain what data they retrieve and under what form
  - [x] Make sure the asynchronous functions are well commented and explained


app/src/main/java/com/lastaoutdoor/lasta/data/api/ActivityRepositoryImpl.kt
```diff
@@ -0,0 +1,276 @@
+package com.lastaoutdoor.lasta.data.api
```
npfrei, line 1: Warning : all of your calls assume the body of the call isn't empty. That can be dangerous in case of error

npfrei (approved):
Amazing job on the API calls! It seems to be finally working great. As you can see in my comment, please check the case where the response's body is empty. Otherwise just be careful when you merge, you modified a lot of files.

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/MapViewModel.kt
```diff
@@ -58,51 +57,40 @@ constructor(private val outdoorActivityRepository: OutdoorActivityRepository) :
...
-    val climbingThread = Thread {
-      climbingNodes =
-          repository
-              .getClimbingActivitiesNode(
-                  rad.toInt(), // in meters
-                  centerLocation.latitude,
-                  centerLocation.longitude)
-              .elements
-    }
+    viewModelScope.launch {
```
JoniDutoit, line 63: Nice! I also thought that we could possibly change the Thread call to a viewModelScope Launch

JoniDutoit (approved):
Great job ! Good refactoring/renaming overall of the project and now we will have more smooth API calls, really nice ! Everything looks good to me, one thing to note though is now that API calls are asynchronous it will possibly be harder to test them but we'll figure it out :)

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/MapViewModel.kt
Andy130604: Thank you, I mean most of your UI had already implemented text that I could just insert the data in. It was pretty easy to integrate thanks to your upfront work!

PR #168, opened 2024-04-24
Language support

npfrei:
Complete translation of the app into both french and german. Strings are stored in values/strings.xml, values-fr/strings.xml and values-de/strings.xml for each respective language. The language of the app is set to the System's default language. If the system's language in not one of french, german or english, it is set to english.
Here are example of the profile page translated each language
image
image
image


ecornamu (approved):
Good job, this was a much needed change :) we still have to decide when and how the user gets to change the `language` of the app, but this PR covers the ressource side. 
Make sure to notice everyone to go this path when creating a `@Composable` that implies words, instead of just hard-coding the values

PR #171, opened 2024-04-24
SettingsScreen on ProfileScreen and SetupScreen only for SignUp. 

### LINKED ISSUE #145: drawer on profile screen for settings

Thimphou:
## Objectives

This PR should close #145 and #170.

## Description

- Implemented the SettingsScreen which we can access via the ProfileScreen (see video below). We need to uncomment the database calls to store the preferences in the UserDB, but for the moment the user is null, but I think Andrew is working on it. 
- Made sure that the SetupScreen was now appearing only for the first-login of a particular user. 

### Authentication and separation SignUp/SignIn
Modified the AuthViewModel and the AuthRepository to observe the SignUp value, and show the screens accordingly. 

### SettingsScreen access
Modified the ProfileScreen to access the screen accordingly. 

### SettingsScreen buttons

- Users can now change their prefered language for the app, and their prefered sport between Hiking and Climbing. 
- Sign out button is now here. 
- Users can also delete their account entirely with the "Delete Account" button. It clears the user's preferences, and everything relating to the user in our DB (still need to uncomment this line because of a null user for the moment). 

https://github.com/LASTA-OUTDOOR/lasta/assets/91149580/184711ab-5a1e-477e-9dca-86a4222c7f24


CassioManuguerra (approved):
Hi Maxime, 
your PR features a nice new setting screen, the code is working great, the tests seem good although there could be a bit more instrumentation tests on your screen.
Some of your lines stay a bit under-commented, as for example in `app/src/main/java/com/lastaoutdoor/lasta/data/auth/AuthRepositoryImpl.kt`, where we can understand your code by reading it thoroughly, but we would benefit from a single comment identifying its purpose ^_^
For suggestions on your solid code one could possibly say that the UI for the setting screen could potentially be revisited later, or think about adding more parameters in `app/src/main/java/com/lastaoutdoor/lasta/data/model/user/UserPreferences.kt`, to have more data stored on the user, like a short description/bio for example.
Good job on your work ✨Thimphou✨ !

PR #174, opened 2024-04-25
Fix small bug for the map screen

JoniDutoit:
## Overview

Quick PR for fixing the display of the map in the Discover Screen that had a hidden recenter button and also refactor package name from "discovery" to "discover"


ecornamu (approved):
Small PR that looks good to me, nothing more to say :3

PR #175, opened 2024-04-25
Weather Report popup dialog for Lausanne

ecornamu:
Implemented an API process that uses `OpenWeatherMap` to display the current weather of a location
- The feature takes place on the `Favorites` tab for now, it was implemented as such to be displayable for tomorrow's demo session but it will be placed somewhere else when we discuss this matter as a team
- For similar reasons, there is no styling on the dialog box yet as it is obviously bound to be decided soon by the team. This PR only includes pure backend work but no UI.
- The query is set to get the weather report for Lausanne, but we can easily change it to display it for the location of an activity or the user's current location
- For now the query only fetches temperature, weather description and wind speed, but the API provides tons of other data that I can easily add if necessary
- Every string is referenced in the `values/strings.xml` ressources and translated in English, French and German
- When merging this to main, don't forget to add the API key to your `local.properties` on your local project but not on the git ! It should be taken care of by the `.gitignore` but still make sure of it so that the key doesn't leak

image
image


Andy130604 (approved):
This is some nice work, you used well the api to get the weather info. Seems easy to extend if wanting to get more weather info. Nonetheless, I could say that you should try to document your code better for other to understand what you've done

PR #180, opened 2024-04-26
Reformat composables to not have viewmodel instances passed for construction

Andy130604:
### Code Implementation
- [x] Implement correct data classes for all the app and db models
- [x] Refactor the navigation to use shared viewmodels with navgraphbuilder construction pattern
- [x] Remove all viewmodels as parameters for composable screens
- [x] Create the appropriate viewmodel states to be used for different screens
- [x] Pass states and lambdas as viewmodel calls and values to the screens


JoniDutoit (approved):
Huge refactor ! Excellent job Andrew, this refactor will enhance significantly the quality of our code.

npfrei (approved):
Reviewed all the files relating to API calls and everything seems good. You really did an amazing job, congrats !

ecornamu (approved):
We cannot thank you enough for the amount of work you put in this PR, this is humongous.
As decided orally for the review split-up during today's meeting, I reviewed every file related to the `Weather Report` and the `More Info` tab and everything seems correct :) well done !!!

Thimphou:
This is some tremendous work that you've done Andrew. This will enhance our testing a lot, and beneficiate everyone to be easily consistent with the database schemes. 
On my side I reviewed every file concerning the authentication, the database and its attributes (preference schemes...), that looks terrifyingly good. 
Thank you a lot :)

CassioManuguerra (approved):
This humonguous PR is really something, nice refactoring, the end-to-end tests will work great with these changes whereas it was complicated before. On the database side, great work too, you modularized really well and it feels really smooth ! 

It is problematic that it created a few bugs in the app, but a PR touching this many files is hard to be completely free from those, and as you already started fixing some of them (the API calls) it should go quickly. On the commentary aspect, it's a bit light but this is normal since it's not your job and we will make sure to add some when we modify the particular parts. Really good job Andrew, it is really a huge workload. 

I would strongly suggest and note though that this should be the last refactor this big that we perform unless it is an absolute necessity, as this (crazy good) work prevented a lot of us from working, and makes everything harder.  

But again, a thousand hurrahs and congratulations !

PR #189, opened 2024-04-30
Made the More Info screen responsive, commented and tested it

CassioManuguerra:
# PR
## Changes
This PR changes the structural implementation of the more info screen. Its UI is no longer is hard coded and is now relative to the screen size and orientation. The start button is now in a Column with the rest in a LazyColumn so that the button appears at all time at the bottom, while preserving correct visibility. The whole more info screen is now tested as well.
Also added comments everywhere in the more info screen.

vertical_more_info
horizontal_more_info

https://github.com/LASTA-OUTDOOR/lasta/assets/91049597/5b74fda8-257f-45e6-a6ba-f91bee46aeca

### Files Changed
- `app/src/androidTest/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreenTest.kt`
- `app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt`

## What's next
- Implement the remaining informations (such as itinerary) to be shown in the more info from database (related to tracking epic -> Milestone 3)
- Maybe un-implement the horizontal mode for Lasta, which is for now supported here (but not in a few parts of the code like the map which is looking awful (while still working technically))


npfrei (approved):
L
       L G
   L  G T
L G T M
   T M
   M

PR #190, opened 2024-05-01
Refactor/map screen

npfrei:
Refactored the MapScreen to remove the MapViewModel as a parameter


ecornamu (approved):
Good refactoring, seems adequate with the new model of the project

JoniDutoit (approved):
Nice work ! Me and Andrew think it looks good and you understand correctly the concepts of why we made the refactor, great job !

PR #193, opened 2024-05-01
Continued the Weather Report Implementation and made app logo visible on phone

ecornamu:
- The app can now query a weather report using latitude and longitude in addition to the already implemented query that used a city name. This second query also uses `OpenWeatherMap API`
- The weather report can now be display in two forms : 

1.  `WeatherReportBig` is the complete report with an icon, location name, humidity %, wind speed and temperature in Celcius. You choose wether you display the wind speed or not using the `displayWind` boolean. This `Composable` is displayed in the `More Info` screen and through a Dialog on `Discovery` screen when you click on the Weather logo 

image
image
2. `WeatherReportSmall` is only comprised of the weather icon that varies according to api response, and the temperature in Celcius. It is displayed on top of the `Discovery` screen. The weather icon is clickable and executes a lambda passed as an argument.

image

- Since the API call needs the user location, the Weather display is unavailable until the user authorizes the app to access their location. On `WeatherReportBig`, a text explaining this is displayed.

image
- I also made the app logo visible on the actual app icon on the main screen of the phone and on app startup

image
image


ecornamu:
As a side remark on why code coverage is below 80%, it is comprised of code that I slightly modified and had no coverage before my fork. It is due to the refactor earlier this week and will be taken care of before the MS2

npfrei (approved):
Great Job ! It seems to works well. The only critique I would give is to commentate your code better. You could have also added some tests to moreInfoScreen since you display the weather within that, but otherwise the testing is great.

CassioManuguerra (approved):
Great job Emile ! This is really nice it looks good works well, you really managed to get the most of this API !

I would strongly suggest to really try to add more comments to your code though, because talking from experience when you come back on somebody else's code which is not commented it is a nightmare. Think ahead, comment your code <3.

Great job, maybe for later we could add weather predictions if we find an easy way but for now this is perfectly fine !

app/src/main/java/com/lastaoutdoor/lasta/ui/navigation/MainNavGraph.kt
Thimphou, line 1: This file clearly takes into account the refractor, thank you for your flexibility on this topic Emile.

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/WeatherViewModel.kt
Thimphou, line 1: ViewModel is a good choice of implementation here, you englobed tricky part of codes with try/catches, that is a good way of proceeding!

ecornamu:
> Great job Emile ! This is really nice it looks good works well, you really managed to get the most of this API !
> 
> I would strongly suggest to really try to add more comments to your code though, because talking from experience when you come back on somebody else's code which is not commented it is a nightmare. Think ahead, comment your code <3.
> 
> Great job, maybe for later we could add weather predictions if we find an easy way but for now this is perfectly fine !

Hello Cassio :) Unfortunately weather prediction is a feature that is not free on the API we use for Lasta.

ecornamu:
I'm merging this PR to main because we are a bit short time-wise but I will comment it later

Thimphou (changes requested):
Hey Emile! 

Your work looks great, and your work will very much beneficiate our app and the way we thought users will scroll through it. Weather is a great add, since we are talking about outdoor sports (Hiking, Climbing and Biking). 

However, I have one tiny change to request: in your `WeatherReport.kt` file, please implement an `else` branch when the `weather` variable is `null`, to make sure your implementation is consistent, does not crash and is fully covering all the possible cases. 
Maybe you could have an `unknown` logo that is returned (to make the offline mode developer easier!). 

This is clearly quick to fix, and I think it will help us :)

On a global note, it is some great work!

app/src/main/java/com/lastaoutdoor/lasta/ui/components/WeatherReport.kt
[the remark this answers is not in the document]
ecornamu: The Composable is then just blank. It's a design choice since this small report is shown on the discovery screen. In case the user didn't accept to share their location, no weather can be displayed

PR #196, opened 2024-05-01
More info/go to map

npfrei:
Allows user to click on the "go to Map" button in the moreInfo tab. I could not really test what I added in the MoreInfoScreen, as it relies on a map tab that's almost impossible to Test. But, I added tests for the rest of the moreInfo screen to see that everything was displayed properly as well as tests for the moreInfoScreenviewModel.


app/src/main/java/com/lastaoutdoor/lasta/ui/navigation/MainNavGraph.kt
CassioManuguerra, line 1: There are a lot of parameters passed to instance a MoreInfoScreen now, maybe it is possible to remove some that are not mandatory ?

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/map/MapScreen.kt
```diff
@@ -191,18 +191,28 @@ private fun GoogleMapComposable(
...
+    updateSheet: (Boolean) -> Unit,
 ) {
   GoogleMap(
       modifier = Modifier.fillMaxSize(),
       properties = state.properties,
       uiSettings = state.uiSettings,
       cameraPositionState = cameraPositionState,
       onMapLoaded = {
+        if (state.selectedMarker.value != null) {
+
```
CassioManuguerra, line 207: This block would look better commented ;-)

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/MoreInfoScreenViewModel.kt
```diff
@@ -26,4 +35,30 @@ class MoreInfoScreenViewModel @Inject constructor() : ViewModel() {
   fun changeActivityToDisplay(activity: Activity) {
     activityToDisplay.value = activity
   }
+
+  fun goToMarker(activity: Activity): com.lastaoutdoor.lasta.models.map.Marker {
```
CassioManuguerra, line 39: It would be nice if you could comment the goToMarker function !

CassioManuguerra (changes requested):
Hello Nehemie, 

your PR is really good the button works fine and the code is neat, although I'm a bit perplexed because when I run your code it seems that all the activities, when you press viewOnMap, take you to a location south of Ghana in the Atlantic Ocean. If it is not the case on your end, maybe you could put some indications on how you can obtain a correct result ?

Apart from this and some functions that are lacking a bit of commenting, you did some great job, handling the map is not easy (🥶) and you managed to do so pretty well !
A few suggestions/notes for next updates (if we have time for it ^^) : 
- Show the other markers around the selected one that is being displayed on the map even if it is not the one that was opened under more info, because if the user is looking for this activity, he/she could be interested in seeing other activities around !
- The UI for the dialog popping under the map will maybe have to be revisited quite a bit, as for now it is not looking great

app/src/main/java/com/lastaoutdoor/lasta/ui/navigation/MainNavGraph.kt
npfrei: Hi, thanks for your comment ! The issue with the points appearing in the middle of Ocean comes from the refactoring of database fuction, so it is not part of my PR. It will be fixed in Andrew's next one.

npfrei: > There are a lot of parameters passed to instance a MoreInfoScreen now, maybe it is possible to remove some that are not mandatory ?

I don't think it is possible without changing (again lol) the architecture of the project, as the mapScreen needs a lot of arguments to be created correctly, but I could look into it in another sprint

CassioManuguerra: Okay thanks ! I am reassured

CassioManuguerra (approved):
It's good ! LGTM

PR #199, opened 2024-05-02
SettingsScreen and SetupScreen adapted with UserLevels, consistent with DB. 

### LINKED ISSUE #177: Test for the SetupScreen and SettingsScreen (EXCEPTIONAL DUE TO REFRACTORING)

Thimphou:
## Objectives

This PR aims to close #177.

Both `SettingsScreen` and `SetupScreen` are now consistent with DB, and taking into account the new sport`Biking`. Users can also select their levels on each of the three sports between `BEGINNER`, `INTERMEDIATE` and `ADVANCED`. 

### Testing

Those two files are tested, with 89% and 91% of line coverage for `SetupScreen` and `SettingsScreen` respectively. 

### Video

Here is a little video of the new SettingsScreen (SetupScreen is quite similar except for the `Sign out` and `Delete account` buttons, which is only `Finished` in SetupScreen. 

https://github.com/LASTA-OUTDOOR/lasta/assets/91149580/816f1cfa-c7d1-40c4-96a8-f5c15adf7082


app/src/main/java/com/lastaoutdoor/lasta/ui/navigation/LoginNavGraph.kt
```diff
@@ -42,11 +46,27 @@ fun NavGraphBuilder.addLoginNavGraph(navController: NavHostController) {
...
+              .value
       SetupScreen(
+          language = language,
+          prefActivity = prefActivity,
+          levels = levels,
           userId = userId,
           updateFieldInUser = authViewModel::updateFieldInUser,
           updateLanguage = preferencesViewModel::updateLanguage,
           updatePrefActivity = preferencesViewModel::updatePrefActivity,
+          updateClimbingLevel = preferencesViewModel::updateClimbingLevel,
```
Andy130604, line 67: This is a nice way of using preferences directly on the setup screen.

Andy130604 (approved):
Overall really nice implementation. You have a lot of same code over setup screen and settings screen, you could maybe think about a way to only have one for both and display different info accordingly

PR #200, opened 2024-05-02
More Info screen improvement

ecornamu:
Changed the UI design of the `More Info` screen to comply with figma, including a system to display the rating with empty, semi-filled and filled stars
WhatsApp Image 2024-05-02 at 15 49 51


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
```diff
@@ -52,7 +52,7 @@ fun MoreInfoScreen(
       // contains the top icon buttons
       item { TopBar(navigateBack) }
       // displays activity title and duration
-      item { ActivityTitleZone(activityToDisplay) }
+      item { ActivityTitleZone(activityToDisplay, processDiffText) }
```
Andy130604, line 55: You don't need to pass lambda, you can use the enum defined functions, either .name or toString()

Andy130604 (changes requested):
Overall really nice design, I just one small remark: you can use enum functions to display text, you shouldn't pass lambda all the way down.

PR #230, opened 2024-05-06
Find small bugs in the app's MS2 version and fix them.

Andy130604:
## Tasks:
- [x] Make a loading/waiting animation for when activities are getting fetched on the discover screen
- [x] Do the same thing for Favorites screen
- [x] Make sure the Search bar color is correct
- [x] Fix the ordering by displaying 6 times the same text
- [x] Fix navigation bug on main navigation graph

### DIFF
--- a/app/src/main/java/com/lastaoutdoor/lasta/ui/components/LoadingAnim.kt
+++ b/app/src/main/java/com/lastaoutdoor/lasta/ui/components/LoadingAnim.kt
@@ -1,0 +1,18 @@
+package com.lastaoutdoor.lasta.ui.components
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.width
+import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun LoadingAnim(width: Int, tag: String = "LoadingBar") {
+  Box(modifier = Modifier.fillMaxSize().testTag(tag), contentAlignment = Alignment.Center) {
+    CircularProgressIndicator(modifier = Modifier.width(width.dp))
+  }
+}
--- a/app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/DiscoverScreen.kt
+++ b/app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/DiscoverScreen.kt
@@ -206,29 +206,6 @@ fun HeaderComposable(
   // Dropdown menu boolean
   var showMenu by remember { mutableStateOf(false) }
 
-  // text for the sorting value
-  val sortingText =
-      when (orderingBy) {
-        OrderingBy.DIFFICULTYASCENDING -> {
-          LocalContext.current.getString(R.string.difficulty_asc)
-        }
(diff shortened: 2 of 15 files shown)


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/DiscoverScreen.kt
```diff
@@ -114,31 +116,37 @@ fun DiscoverScreen(
...
-                changeActivityToDisplay,
-                flipFavorite,
-                navigateToMoreInfo)
+          Spacer(modifier = Modifier.height(8.dp))
+          if (isLoading) {
+            LoadingAnim(width = 35, tag = "LoadingBarDiscover")
+          } else if (activities.isEmpty()) {
+            /* TODO */
+          } else {
+            LazyColumn {
```
JoniDutoit, line 139: Great job! It's indeed better to have the lazy column being just the list of activities and have a fixed header, wouldn't have thought of it myself !

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/favorites/FavoritesScreen.kt
```diff
@@ -1,35 +1,64 @@
...
+  Column(modifier = Modifier.testTag("FavoritesScreen")) {
+    Box(
+        modifier = Modifier.fillMaxWidth().height(70.dp),
+        contentAlignment = Alignment.CenterStart) {
+          Text(
+              text = LocalContext.current.getString(R.string.tab_favorites),
+              style = MaterialTheme.typography.headlineLarge,
+              modifier = Modifier.padding(start = 20.dp))
+        }
+    HorizontalDivider()
```
JoniDutoit, line 44: Great reutilization of already existing composables!

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/favorites/FavoritesScreen.kt
```diff
@@ -1,35 +1,64 @@
...
+    if (isLoading) {
+      LoadingAnim(width = 35, tag = "LoadingBarFavorites")
+    } else if (favorites.isEmpty()) {
```
JoniDutoit, line 47: If it's empty, we could simply display some text, nice!

JoniDutoit (approved):
Excellent job! Navigation looking really good and clean, nice improvements done for the application. It's cool to see how the app's is building up. I wonder if we could do the same loading feature for the MapScreen as well, maybe only showing the map when it's completely loaded up could be a nice idea? We can discuss it over the next meeting, keep the good work!

app/src/main/java/com/lastaoutdoor/lasta/ui/components/LoadingAnim.kt
```diff
@@ -0,0 +1,18 @@
...
+import androidx.compose.foundation.layout.width
+import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun LoadingAnim(width: Int, tag: String = "LoadingBar") {
```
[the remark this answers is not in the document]
Andy130604: The test tag is different since it is displayed on different screens and the tags need to be unique in the case of and E2E test. Also the width param is the one for the circular progress indicator and that can be different for instance on the loading screen it is quite large but on the discover screen can be smaller to emphasise that its loading a subset of the app's content.

PR #233, opened 2024-05-07
Start the tracking feature of the app. Implement a stopwatch/timer and have gps signal updated and returned.

Andy130604:
## Tasks:
- [x] Build a stopwatch bound to the android services and display it on the tracking screen
- [x] Make GPS update calls every ... seconds
- [x] Get GPS location data everytime we make the update call and start forming a list of data points
- [x] DISCUSS: How to store these in DB and when to make these calls.
- [x] Make sure everything is well documented
- [x] Make sure everything is well tested

Sonar is really buggy 2 whole files are tested but display 0% coverage.


npfrei (approved):
LGTM

PR #234, opened 2024-05-08
Made the weather report relative to the currently displayed activity's location 

ecornamu:
- Modified the `WeatherActivityViewModel` so that you can switch between weather for the current user location and weather for target location in latitude/longitude passed as arguments. The app therefore displays the current location weather report on the `Discover` tab, and the weather on the `More Info` screen is from the displayed activity location.
(FYI, `Montain View` is the name of the city where my emulator seems to be located, as silly as it sounds)

image

image

- The `WeatherActivityViewModel` is now also a shared view model so that the current weather isn't related to the `Composable` that builds it

- Tried to add test suites to `WeatherActivityViewModel`, but the need of an `Application` that validates that the user authorized location usage seems to make it difficult and I didn't achieve it for now. Maybe this is why the already present test is commented ? To compensate for this lack of coverage, I added tests for `WeatherRepositoryImpl` and additional coverage to `MoreInfoScreen`

### DIFF
--- a/app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
+++ b/app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
@@ -69,18 +69,15 @@ fun MoreInfoScreen(
 ) {
   var isMapDisplayed = remember { mutableStateOf(false) }
   if (!isMapDisplayed.value) {
-    Column(modifier = Modifier
-        .fillMaxSize()
-        .testTag("MoreInfoComposable")) {
-      LazyColumn(modifier = Modifier
-          .weight(1f)
-          .padding(8.dp)) {
+    Column(modifier = Modifier.fillMaxSize().testTag("MoreInfoComposable")) {
+      LazyColumn(modifier = Modifier.weight(1f).padding(8.dp)) {
         item { Spacer(modifier = Modifier.height(15.dp)) }
         // contains the top icon buttons
-        item { TopBar {
+        item {
+          TopBar {
             navigateBack()
             setWeatherBackToUserLoc()
-        }
+          }
         }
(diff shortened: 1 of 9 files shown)


app/src/test/java/com/lastaoutdoor/lasta/data/api/weather/WeatherRepositoryImplTest.kt
Thimphou, line 1: These tests look good, maybe you could try mockking the application to achieve more test coverage?

Thimphou (changes requested):
Hey Emile! 

That is some nice work overall, I'm just requesting changes for the typo which is fixable in 10 seconds, otherwise this is a great feature for our users :)

app/src/test/java/com/lastaoutdoor/lasta/data/api/weather/WeatherRepositoryImplTest.kt
ecornamu: The problem is deeper than that unfortunately, the `WeatherViewModel` verifies that the user location usage is allowed. This behaviour is weird to simulate with `mockk` and I ultimately didn't achieve it.

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
```diff
@@ -63,16 +63,22 @@ fun MoreInfoScreen(
     goToMarker: (Activity) -> Marker,
     weather: WeatherResponse?,
     markerList: List<Marker>,
-    selectedItenary: MapItinerary?,
+    selectedItinary: MapItinerary?,
```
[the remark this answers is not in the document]
ecornamu: My bad, fixed now :)

Thimphou (approved):
Thanks :)

PR #235, opened 2024-05-08
Discovery Screen - Autcompleted searchbar

JeremyHugentobler:
# Searchbar autocomplete

## Content
When writing in the Discover Screen's searchbar, known cities / location are suggested beneath the searchbar.
If you click on a location, the discover screen changes it location to the suggested places and the displayed activities are updated.
The same goes for the map. After a click on a suggestion, the camera is moved to the right place and activities updated.

## API
In order to obtain correct suggestions, I used the [Radar API](https://radar.com) that returns suggestions of places when provided with a start input. We have a 2 week enterprise free trial, but after this period of time, the amount of query will be limited (100k per day).

The API needs a key and therefore you will all need to update your local.properties file !

## Next
- Still a small UX incovenience due to the keyboard focus after clicking on a suggestion, but I cannot figure how to fix this for the moment

## Illustrations
image
image


app/src/main/java/com/lastaoutdoor/lasta/data/api/autocomplete/RadarRepositoryImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+import javax.inject.Inject
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.suspendCancellableCoroutine
+import retrofit2.Call
+import retrofit2.Callback
+
+class RadarRepositoryImpl @Inject constructor(private val radarApiService: RadarApiService) :
+    RadarRepository {
+
+  private val layers: String = "county,locality,state"
```
JoniDutoit, line 15: is there a typo or is it supposed to say "county" instead of "country"? Also, could make this a constant!

app/src/main/java/com/lastaoutdoor/lasta/models/api/RadarData.kt
```diff
@@ -0,0 +1,25 @@
...
+    @SerializedName("longitude") @Expose val longitude: Double,
+) {
+
+  // returns the label of the suggestion
+  fun getSuggestion(): String {
+    return label
+  }
+
+  // returns the position of the suggestion
+  fun getPosition(): LatLng {
```
JoniDutoit, line 22: Nice utilization of already existing methods!

app/src/test/java/com/lastaoutdoor/lasta/data/api/autocomplete/RadarRepositoryImplTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+class FakeRadarApiServiceTest : RadarApiService {
+  override fun getSuggestions(
+      query: String,
+      layers: String,
+      country: String,
+      limit: Int
+  ): Call<RadarApiResponse> {
+    return mock(Call::class.java) as Call<RadarApiResponse>
+  }
```
JoniDutoit, line 15: Be careful to use mockito library instead of mockk, now we're mixing two mocking libraries amongs our tests

JoniDutoit (dismissed):
Nicely done! Such a cool feature to have in the application, working&looking great.

JoniDutoit (approved):
Nice! We can merge with main now

app/src/main/java/com/lastaoutdoor/lasta/data/api/autocomplete/RadarRepositoryImpl.kt
JeremyHugentobler: No, it is not a typo, the county is the equivalent of "Cantons" (the API was developed in the US so they just translated that to this I think)

PR #236, opened 2024-05-08
Improve Discover Screen UI and fix DropDownMenu component

JoniDutoit:
## Overview
This PR aims to create a homogenous feeling to the application, hence making all DropDownMenu equals and responsive to the Language System, as well as minors UI fixes for the Discover Screen. 

## Tasks:
- [x] Adapt the DropDownMenuComponent to display strings from the Resources (R.string.xxx) 
- [x] Fix all screens that use this DropDownMenu (Settings, Discover, Profile) 
- [x] Small UI improvement on the Weather for the Discover 
- [x] Test new code

### Discover
Before | After
---------- | -----------
 |

### Settings/Setup
Before | After
---------- | -----------
 |

### Profile
Before | After
---------- | -----------
 | 
## Testing
Achieved 100% line coverage on new code


ecornamu (dismissed):
- Good job giving a fresher look to the app, you are really our head of design lol. Maybe include `screenshots` so that everyone can see what the newer designs are. You showed me the changes IRL so I know what I'm looking at but that's not the case for everyone ;) It would be silly to request changes for that but think about it for next time !
- I also see that you put additional ressources in the `strings.xml ` and corrected some syntax and I thank you for it, some stuff was just unaesthetic and needed change

ecornamu (approved):
Only tests since my last review, still approved :)

app/src/main/java/com/lastaoutdoor/lasta/models/activity/ActivityType.kt
JeremyHugentobler, line 1: Clean restructuration thank you

JeremyHugentobler (approved):
Hi,

You did a good job restructuring some parts of the code (enums) !
The modifications of the UI that you provided are a good step towards the UI "unification" of our app.

There are still to much components created by each one of us that could be regrouped to a single one that would give a more coherent look to the app, but this is a nice start and we should think about it everytime we implement something new.

Thank you for your work !

PR #240, opened 2024-05-09
Added functions to the UserActivities repository and tested the whole

ecornamu:
- Added utilitary functions to the `UserActivitiesDBRepositoryImpl`, such as `getNLatestActivities(userId: String, n: Int)` to get the n activities with the latest `timeFinished`
- Tested the added functions and the already implemented ones for a line coverage of 95% according to supreme chancelor Jacoco

image


CassioManuguerra (changes requested):
Hey Emile,
your PR is very concise and effective, and those functions will surely come in handy, thanks for writing them !  You commented strongly the function headers in the interface, which is a great thing ! 
Your tests are quite solid too, nice job on getting all the coverage :-).

Because we have not used these functions in our code yet, even if they work with the unit tests doesn't mean they work properly though, and it would be good to add a small comment on them saying something like "may cause small bugs", even if they are pretty straight-forward, in case we have trouble later with the database.
An exemple of what I'm saying is that I see one potential small problem with your `getNLatestActivities` in `UserActivitiesDBRepositoryImpl.kt`, and that is that you take(n) without checking if there are at least n elements in the list, I think that take() from kotlin will just return as much as element possible but it would still be good to check edge cases like this when writing these functions because we want to avoid as much as possible a crash case ^^

Apart from that great job, I would suggest maybe writing more activity DB functions later because there are many other parameters that we can filter as you did with the ActivityType, but we'll see if we need more when we get there !

CassioManuguerra:
> > Hey Emile, your PR is very concise and effective, and those functions will surely come in handy, thanks for writing them ! You commented strongly the function headers in the interface, which is a great thing ! Your tests are quite solid too, nice job on getting all the coverage :-). They could maybe be tested a bit more with actual fake values but it's perfectly fine as it is.
> > Because we have not used these functions in our code yet, even if they work with the unit tests doesn't mean they work properly though, and it would be good to add a small comment on them saying something like "may cause small bugs", even if they are pretty straight-forward, in case we have trouble later with the database. An exemple of what I'm saying is that I see one potential small problem with your `getNLatestActivities` in `UserActivitiesDBRepositoryImpl.kt`, and that is that you take(n) without checking if there are at least n elements in the list, I think that take() from kotlin will just return as much as element possible but it would still be good to check edge cases like this when writing these functions because we want to avoid as much as possible a crash case ^^
> > Apart from that great job, I would suggest maybe writing more activity DB functions later because there are many other parameters that we can filter as you did with the ActivityType, but we'll see if we need more when we get there !
> 
> Hi Cassio :) As stated in this article from the kotlin academy, the case when the lists contains <n elements is handled by the take function and therefore doesn't crash.
> 
> image https://kt.academy/article/fk-cp-drop-take

Okay great ! Then it's fine for now :) I would still request you to add a small comment `//correct functioning to be checked with main code ` and then all good for me !

ecornamu:
> > > Hey Emile, your PR is very concise and effective, and those functions will surely come in handy, thanks for writing them ! You commented strongly the function headers in the interface, which is a great thing ! Your tests are quite solid too, nice job on getting all the coverage :-). They could maybe be tested a bit more with actual fake values but it's perfectly fine as it is.
> > > Because we have not used these functions in our code yet, even if they work with the unit tests doesn't mean they work properly though, and it would be good to add a small comment on them saying something like "may cause small bugs", even if they are pretty straight-forward, in case we have trouble later with the database. An exemple of what I'm saying is that I see one potential small problem with your `getNLatestActivities` in `UserActivitiesDBRepositoryImpl.kt`, and that is that you take(n) without checking if there are at least n elements in the list, I think that take() from kotlin will just return as much as element possible but it would still be good to check edge cases like this when writing these functions because we want to avoid as much as possible a crash case ^^
> > > Apart from that great job, I would suggest maybe writing more activity DB functions later because there are many other parameters that we can filter as you did with the ActivityType, but we'll see if we need more when we get there !
> > 
> > 
> > Hi Cassio :) As stated in this article from the kotlin academy, the case when the lists contains <n elements is handled by the take function and therefore doesn't crash.
> > image https://kt.academy/article/fk-cp-drop-take
> 
> Okay great ! Then it's fine for now :) I would still request you to add a small comment `//correct functioning to be checked with main code ` and then all good for me !

It is done my dear

CassioManuguerra (approved):
Great thanks a lot !

PR #241, opened 2024-05-09
Added activities filtering on the discovery screen

ecornamu:
Now the activities displayed on the Discover screen are relative to what the user puts on the Filter screen, i.e. the `ActivityType` and the `UserLevel` difficulty the user wants to see in the list


app/src/main/java/com/lastaoutdoor/lasta/viewmodel/DiscoverScreenViewModel.kt
```diff
@@ -110,7 +111,7 @@ constructor(
     }
   }
 
-  private fun activitiesToMarkers(activities: List<Activity>): List<Marker> {
```
JoniDutoit, line 113: is it normal to change this method from private to public? Is that the desired result? Because if we're not calling this function in any other place, maybe we should keep it private :)

JoniDutoit (commented):
Good job! This helps as for the beginning of a more dynamic/responsive discover screen, soon it will be working perfectly 💯

JoniDutoit (approved):
Besides my comment, everything looks good!

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/DiscoverScreenViewModel.kt
ecornamu: Yeah I hesitated but it's just so that I'm able to test it in `DiscoveryScreenViewModelTest` to provide additional line coverage. It's weird but a private function cannot be tested BUT counts as uncovered lines

PR #242, opened 2024-05-09
Community - Message UI changes 

JeremyHugentobler:
## Context
Previously, I could not find a way to create a standard message UI with  a textfield at the bottom of the screen, because the keyboard appeared over it.

## Content
I removed the old "form" to send messages and replaced it with a more standard UI for a chatting screen.
I also changed the color and max width of messages in order to make the page clearer.

A message can be written on multiple lines and if you click somewhere else on the page, the keyboard is dismissed.

## Illustrations
[Screen_recording_20240509_150503.webm](https://github.com/LASTA-OUTDOOR/lasta/assets/94183393/8bebd7e5-2a16-468d-bbfc-6ba372153c56)

Difference of colors (sending vs receiving)
image


Andy130604 (approved):
Everything looks great, nice implementation can't wait to send you messages through this new interface

PR #243, opened 2024-05-09
OfflineMode

npfrei:
Implemened an offlineMode for activities , allowing user to save activtities locally using the download  button in the MoreInfo tab. Users can then see these activities in the favorites tab while offline


app/src/main/java/com/lastaoutdoor/lasta/models/user/Converters.kt
Thimphou, line 1: Great way of doing so!

Thimphou (approved):
Your code looks good, thank you for your implementation! However, please make sure to comment in the upcoming future your functions in the DAO files!

Thimphou (approved):
Nice testing that improves our code coverage!

app/src/main/java/com/lastaoutdoor/lasta/models/user/Converters.kt
```diff
@@ -0,0 +1,96 @@
...
+  @TypeConverter
+  fun fromClimbingStyle(c: ClimbingStyle): String {
+    return c.toString()
+  }
+
+  @TypeConverter
+  fun toClimbingSyle(s: String): ClimbingStyle {
+    return if (s == "Outdoor") {
+      ClimbingStyle.OUTDOOR
+    } else {
```
ecornamu, line 21: Maybe be careful with this else if in the future we add a third climbing style or change the naming

app/src/main/java/com/lastaoutdoor/lasta/models/user/Converters.kt
```diff
@@ -0,0 +1,96 @@
...
+    return d.toString()
+  }
+
+  @TypeConverter
+  fun toDifficulty(s: String): Difficulty {
+    return if (s == "Easy") {
+      Difficulty.EASY
+    } else if (s == "Normal") {
+      Difficulty.NORMAL
+    } else {
```
ecornamu, line 37: Same remark for here, etc

ecornamu (approved):
Good job ! The implementation is really clever and was much needed, thank you for spending time on it for the past weeks :) I added small comments on the `else` cases of your type transitions but they are not worth a change request, it's only something to take into account in case we change some naming conventions or add possibilities to some enums. Maybe a case switch would be better in this context

(Be careful in case you didn't notice the CI fails for the moment but I'm sure you are on it)

ecornamu (approved):

PR #257, opened 2024-05-11
PR: Profile Screen fix

JoniDutoit:
## Overview

This small PR aims to fix the behavior and UI display of the Profile Screen after the refactor. Merge this PR will give the following results as shown below

## Results
Before | After
---------- | -----------
 |


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/profile/ProfileScreen.kt
```diff
@@ -429,89 +435,99 @@ fun Chart(activities: List<UserActivity>, timeFrame: TimeFrame, sport: ActivityT
...
-                          text =
-                              String.format(
-                                  "%.2f", metersToKilometers(climb.totalElevation.toLong())),
-                          fontWeight = FontWeight.Bold,
-                          modifier = Modifier.testTag("TestAndrew1"))
-                      Text(text = LocalContext.current.getString(R.string.elevation))
-                    }
-                    ActivityType.BIKING -> {}
+  Text(
+      "Recent Activities",
```
Andy130604, line 440: Should have this in string resources

Andy130604 (changes requested):
Everything looks good you have one small fix to do: Have recent activities text be in string resources.

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/profile/ProfileScreen.kt
JoniDutoit: Thank you for pointing it out, indeed I didn't see it ! Made the change rn

Andy130604 (approved):
Everything seems fixed!

PR #261, opened 2024-05-12
Only testing PR for the OSM API service

Andy130604:


app/.gitignore
```diff
@@ -1,6 +1,7 @@
 ./idea
 
 /build
+/schemas
```
JeremyHugentobler: folder generated by Android Rooms and indeed it should not be on the git

JeremyHugentobler (approved):
API testing was still a big missing part of the team's testing capabilities, but this changes everything !

We will definitively follow your work for the next tests we are going to write !

PR #262, opened 2024-05-13
UI redesign of the filter screen

JoniDutoit:
## Overview
The Filter Screen didn't follow the same UI model that the rest of the application, hence an UI redesign was necessary in order to give that homegenous feeling to the application. So, this PR aims to resolve this problems, given a new fresh look to the Filter Screen

## Note
This PR does NOT aim to implemente the functionality (backend) of the filter screen. This means that the filter screen should not modify the display nor the ordering of the activities in the Discover Screen if merged with the main. Short history, it's purely an UI feature. 

## Results
Before | After
---------- | -----------
 | <video src="https://github.com/LASTA-OUTDOOR/lasta/assets/95919886/1324a234-3cd5-495a-85dd-3b37d3e908a5" width="200">


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/FilterScreen.kt
```diff
@@ -19,190 +21,207 @@ import androidx.compose.material3.ExperimentalMaterial3Api
...
+        },
+        navigationIcon = {
+          IconButton(onClick = { navigateBack() }) {
+            Icon(Icons.AutoMirrored.Filled.ArrowBack, contentDescription = null)
+          }
+        })
+
+    HorizontalDivider(modifier = Modifier.fillMaxWidth())
+
+    Spacer(modifier = Modifier.height(16.dp))
```
JeremyHugentobler, line 97:
```suggestion
    Spacer(modifier = Modifier.fillMaxHeight(0.05f)
```

With spacer, instead of harcoding a value (In this case it is not a problem since it renders properly on most of devices), you can use properties like Modifier.fillMaxHeight(0.05f) which will take a percentage of the available height (5% here) and therefore will adapt itself with various screen size.

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/FilterScreen.kt
```diff
@@ -19,190 +21,207 @@ import androidx.compose.material3.ExperimentalMaterial3Api
...
+        verticalArrangement = Arrangement.SpaceBetween) {
+          Column {
+            Text(
+                text = stringResource(id = R.string.filter_activity_type),
+                style = MaterialTheme.typography.headlineSmall,
+                fontWeight = FontWeight.Bold)
+
+            Spacer(modifier = Modifier.height(16.dp))
+
+            FlowRow(
```
JeremyHugentobler, line 112: I don't think that the button in the FlowRow are centered /spaced arround correctly. If you look at the vidéo preview, there is more space from the last button to the screen border than from the left screen border to the first button

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/FilterScreen.kt
```diff
@@ -19,190 +21,207 @@ import androidx.compose.material3.ExperimentalMaterial3Api
...
+                verticalArrangement = Arrangement.spacedBy(4.dp),
+                horizontalArrangement = Arrangement.spacedBy(16.dp),
```
JeremyHugentobler, line 114:
```suggestion
                horizontalArrangement = Arrangement.SpaceEvenly(),
```
might resolve the previously mentioned arrangement issue

JeremyHugentobler (changes requested):
Hi, thank you for your work, the screens looks cleaner and more unified with the rest of the app now.

The tests seemed good to me, but there are two UI details that could be improved:
- I left a comment and a suggestion in your PR about the arrangement of the filter buttons. They are not evenly space horizontally and this could be fixed rather easily (more space from the the left border to the first button than from the last button to the right border)
- The color of the small text under the checkbox might be a bit to light and  makes it hard to read on a white background

Additionally,  you could make spacers of dynamic sizes (I left a suggestion in the PR) but this is not really a problematic issue.

Thank you

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/FilterScreen.kt
JoniDutoit: Yes, indeed it could be a problem for smaller screen devices, well pointed out !

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/FilterScreen.kt
JoniDutoit: Yes indeed. I felt that arranging items from a FlowRow harder than usual, as it's constly adapting with n or fewer elements, hence spacing can sometimes be awkward depending on the size of the buttons

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/FilterScreen.kt
JoniDutoit: Yes, I implemented it, it does output a better result, nice suggestion!

JeremyHugentobler (approved):
Thank you very much for your changes !

PR #263, opened 2024-05-13
Stepcounter

ecornamu:
Do not review yet please :)

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -4,3 +4,4 @@
 /app/build
 local.properties
 google-services.json
+/schemas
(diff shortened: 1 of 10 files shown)


npfrei (changes requested):
Great Job ! The way you modularized the composables functions is great. Yet, you only have 54% of code coverage. If it is possible, please add more tests to get to at least 80% (or somewhat close to it)

ecornamu:
> Great Job ! The way you modularized the composables functions is great. Yet, you only have 54% of code coverage. If it is possible, please add more tests to get to at least 80% (or somewhat close to it)

Hi :) As stated above, it is not :/ For now, we need to consider this side as hardly testable as we did for the `MapScreen`

npfrei (approved):
Okay if testing is not possible then everything is good !

PR #264, opened 2024-05-14
Share button for activities

JeremyHugentobler:
# Share Function for Activities
## Summary
This PR introduces a new function that launches an Android sharesheet for sharing activities within our app. The sharesheet includes the app's name and a link to the activity, and supports translations in both French and German. Additionally, the share button on the "More Info" screen now triggers this function.

## Details
- **Functionality** : A new function has been implemented that, given the app context and an activity, launches an Android sharesheet.
- **Sharesheet Content** : The sharesheet shares a text containing the app's name and a link to the activity. The content is translated into French and German.
- **Future enhancements** : Implement a system to open shared links directly within the app. The current share function already includes the link in the shared text, paving the way for this feature.
- **Content modification** : It is rather easy to change the text that is being shared, it should be straightforward to upgrade this functionnality in the future

## Recording
[Sharing PR.webm](https://github.com/LASTA-OUTDOOR/lasta/assets/94183393/211ad813-5c7e-4468-9545-430b0793c2ec)

### DIFF
--- a/app/src/main/java/com/lastaoutdoor/lasta/ui/components/ShareSheets.kt
+++ b/app/src/main/java/com/lastaoutdoor/lasta/ui/components/ShareSheets.kt
@@ -1,0 +1,24 @@
+package com.lastaoutdoor.lasta.ui.components
+
+import android.content.Context
+import android.content.Intent
+import android.text.Html
+import androidx.core.text.HtmlCompat
+import com.lastaoutdoor.lasta.R
+import com.lastaoutdoor.lasta.models.activity.Activity
+
+
+//Opens the android share sheet, will contain a description and a link to the activity -> will be used to share activities
+fun shareActivity(activity : Activity, context: Context) {
+
+    val sendIntent: Intent = Intent().apply {
+        action = Intent.ACTION_SEND
+        putExtra(Intent.EXTRA_TEXT,
+            context.getString(R.string.share_body) + "\n${activity.name} \nhttps://lasta.jerem.ch/activities/${activity.activityId}"
+        )
+        type = "text/plain"
+    }
+
+    context.startActivity(Intent.createChooser(sendIntent, context.getString(R.string.share_title)))
+
+}
--- a/app/src/main/java/com/lastaoutdoor/lasta/ui/components/ShareSheets.kt
+++ b/app/src/main/java/com/lastaoutdoor/lasta/ui/components/ShareSheets.kt
@@ -2,23 +2,22 @@ package com.lastaoutdoor.lasta.ui.components
 
 import android.content.Context
 import android.content.Intent
-import android.text.Html
-import androidx.core.text.HtmlCompat
 import com.lastaoutdoor.lasta.R
 import com.lastaoutdoor.lasta.models.activity.Activity
 
+// Opens the android share sheet, will contain a description and a link to the activity -> will be
+// used to share activities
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/lastaoutdoor/lasta/ui/components/ShareSheets.kt
```diff
@@ -0,0 +1,23 @@
...
+// used to share activities
+fun shareActivity(activity: Activity, context: Context) {
+
+  val sendIntent: Intent =
+      Intent().apply {
+        action = Intent.ACTION_SEND
+        putExtra(
+            Intent.EXTRA_TEXT,
+            context.getString(R.string.share_body) +
+                "\n${activity.name} \nhttps://lasta.jerem.ch/activities/${activity.activityId}")
```
ecornamu, line 18: Really a nitpick but this url bit could also be in a string ressource in case we use it elsewhere

app/src/main/res/values-de/strings.xml
```diff
@@ -108,6 +108,8 @@
     <string name="erase_options">Alles löschen</string>
     <string name="ride">Radtour</string>
     <string name="recent_activities">Neueste Aktivitäten</string>
+    <string name="share_title">Lasta : Aktivitäten teilen</string>
+    <string name="share_body">Hallo ! sieh dir diese Aktivität auf Lasta an :</string>
```
ecornamu, line 112: The capital letter is missing !

app/src/main/res/values-fr/strings.xml
```diff
@@ -108,6 +108,8 @@
     <string name="erase_options">Tout effacer</string>
     <string name="ride">Tour</string>
     <string name="recent_activities">Activités récentes</string>
+    <string name="share_title">Lasta : Partage d\'activité</string>
+    <string name="share_body">Hello ! Regardes cette activité sur Lasta :</string>
```
ecornamu, line 112: The verb "regarder" to the impératif tense is "regarde" without an s at the end :)

ecornamu (changes requested):
Well done ! Few remarks : 

- As stated in the code comments, there are two grammar mistakes in translations that you can fix on a whim. It would also be clever to put the url in a string ressource so that you only need to change it in one place in case the exact url changes. This is why I request those changes for your pr.

- This is not a change I request but rather an idea to remember : in case we need to launch an intent to share something else than an activity, maybe it would be practical to have a more general `shareIntentLauncher` that other devs could reuse ?

app/src/main/res/values-de/strings.xml
JeremyHugentobler: Thank you, it has been corrected

app/src/main/res/values-fr/strings.xml
JeremyHugentobler: Also corrected, thank you

app/src/main/java/com/lastaoutdoor/lasta/ui/components/ShareSheets.kt
JeremyHugentobler: Thank you for the remark, it is indeed a better choice and I will change this immediately

JeremyHugentobler:
@ecornamu Thank you for your review ! The changes requested were made

ecornamu (approved):
Good job ! Much love

PR #268, opened 2024-05-15
Connected filter options to the activity display on Discover screen

ecornamu:
- The filter options you choose in the `FilterScreen` are now persistent and actually modify the displayed `Activities` on the `DiscoverScreen` when you press the `Apply` button at the end of the screen. 

- The `Erase` button now also reverts the changed filtering options back to the previous ones (last time you pressed `Apply`)

- There are a lot of commits on this PR but it was just me fighting against the CI.  It was failing a test that was passing on my machine. I ultimately just deleted the concerned test to stop waisting time, hence the low coverage of 58% which amounts to 13 uncovered lines of code


npfrei (approved):
Looks good, the coverage is somewhat too low but i get that this screen is very difficult to to test

ecornamu:
> Looks good, the coverage is somewhat too low but i get that this screen is very difficult to to test

Actually I can test it, the CI just isn't okay with it for some reason :/

PR #269, opened 2024-05-15
MoreScreen: Making it not hard coded and fixing bugs

Thimphou:
## Objectives

- `MoreInfoScreen` is now modularized to show every component on different screen sizes (following MS2 feedback). I also added default icons for activities when there is no activity images (instead of the previous blue ellipse). 

- When you add a rating on an activity, it is now synchonized with `DiscoverScreenViewModel` and the number of ratings and mean rating are updated. Reduced a few `fetchActivities` calls since some were a bit useless, to improve the app fluidity. 

- Fixed `RangeSearchComposable` range to be remembered when you change screens and you come back, you still have the good range (so the good activities displayed). 

## Testing

I adapted previous tests, thus achieving 76.3% branch coverage but 83.2% LINE COVERAGE which is the one required to be >80%. 

## Video

https://github.com/LASTA-OUTDOOR/lasta/assets/91149580/8dba1f6c-2d71-4f33-ae6f-3fdebfdcff9a


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/components/RangeSearchComposable.kt
```diff
@@ -41,8 +41,9 @@ fun RangeSearchComposable(
     selectedLocality: Pair<String, LatLng>,
     setRange: (Double) -> Unit,
     setSelectedLocality: (Pair<String, LatLng>) -> Unit,
-    fetchActivities: (Double, LatLng) -> Unit,
     isRangePopup: Boolean,
+    updateRange: (Double) -> Unit,
```
Andy130604, line 45: Nice way of having the range be synced across whole screen.

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
```diff
@@ -472,23 +469,27 @@ fun AddRatingButton(
...
+                    style = TextStyle(fontSize = 18.sp, fontWeight = FontWeight.Bold))
 
                 Spacer(modifier = Modifier.height(30.dp))
 
                 Row(
                     horizontalArrangement = Arrangement.SpaceAround,
                     verticalAlignment = Alignment.CenterVertically) {
                       Text(
-                          text = "Please rate (out of 5 stars): ",
+                          text = LocalContext.current.getString(R.string.ask_rating),
```
Andy130604, line 483: Nice change for language support

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/DiscoverScreenViewModel.kt
```diff
@@ -165,25 +165,31 @@ constructor(
     _suggestions.value = emptyMap()
   }
 
-  fun fetchActivities(rad: Double = 10000.0, centerLocation: LatLng = LatLng(46.519962, 6.633597)) {
+  fun fetchActivities() {
     viewModelScope.launch {
       _isLoading.value = true
-      _activities.value = ArrayList()
-      _activityIds.value = ArrayList()
+      val activitiesHolder: ArrayList<Activity> = ArrayList()
```
Andy130604, line 171: You don't need to use temp lists here, you can just update the state value directly.

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/MoreInfoScreenViewModel.kt
```diff
@@ -63,6 +63,7 @@ constructor(
         is Response.Loading -> {}
         is Response.Success -> {
           val osmData = response.data!!
+          println("YOUGOU")
```
Andy130604, line 66: you should remove all prints for a PR

Andy130604 (changes requested):
You can see the individual comments for the changes. Otherwise great job changing the viewmodel fetch activities call. I think this will greatly improve performance on that screen!

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/MoreInfoScreenViewModel.kt
Thimphou: My bad for that, it is now deleted

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/DiscoverScreenViewModel.kt
Thimphou: In fact, I need to do it since in iterate through all the activity types in the for loop. Thus, I need to have a holder list that adds everything at each iteration, if I do what you are suggesting, I might have only the last iteration of the loop and not all the previous ones.

Andy130604 (approved):
Thank you for the changes. I understand about the list, makes didn't have the insights of the development part. Nice testing overall with over 80% line coverage on new code. Great job!

PR #272, opened 2024-05-16
Fix for translation

npfrei:
This a fix for my previous pr, which crashed on older android versions (i.e older than Tiramisu). This is now fixed.


Thimphou (approved):
LGTM

PR #277, opened 2024-05-16
Error Manager implementation for handling errors.

Thimphou:
## Description

This PR aims to catch all calls in the app that could cause the app to crash! In fact, now when we make a call that throws an exception, it is properly catched and it displays a `ErrorToast` that has a personnalized message about the exception occuring.

## Implementation

I adapted every ViewModel and surrounded calls that can fail (API, Database, FirebaseAuthentication, ...), and treated them accordingly. 

## Video

Here, you can see how it appears on our app, when I put myself offline, it shows which calls would have make the app crash.

https://github.com/LASTA-OUTDOOR/lasta/assets/91149580/be1f9347-552b-471f-8883-040a856b4b00


CassioManuguerra (changes requested):
Hi Maxime, your code is really nice ! This is a core feature to have, even if in our case it comes a bit late, but great functionality ! 
Your code is really clean, and it's a tough work for such a feature.
I have to say though that again, the comments are not there. Please add some !
I also see that your coverage is under 80%, It would be nice if you can get it up to at least 80%. 
So for changes please add
- comments
- line coverage

Thank you for your hard dedicated work, your code rocks !

JoniDutoit (changes requested):
Great job! The Toast ideia for messages when there's an error is great, it's used in some applications. That said, I do find it more appealing the Jeremy's approach of the offline mode where informations is not at all displayed and there's a background image error stating that you're indeed offline, maybe could be a future improvement? 

Other than that, your code looks really clean, but unfortunately of yet you haven't achieved 80% line coverage, so I can't approve this PR for now.

Andy130604 (approved):
Overall this PR comes in so handy for reducing the app's number of ways of crashing. I think you could have been more precise with the try catches by catching specific errors but since you handled the try catches very locally, I guess it doesn't change much. Great testing too! Keep up the good work @Thimphou

CassioManuguerra (approved):
Me like it, good job max your PR is looking sweet !
I would suggest if we have time in the final sprint to modularize a bit more, but great thanks for changing what I requested, approving now !

JoniDutoit (approved):

PR #292, opened 2024-05-21
Database functions to delete everything related to a particular user.

Thimphou:
## Description

This PR aims to create database functions to delete a particular user's information from all our collections (`activities`, `conversations`, `user_activities`, `user_fcm_tokens` and `users`.

#### `AuthRepositoryImpl.kt`
- **deleteAccount**: Deletes the current authenticated user's account. This includes removing the user from the `userDBRepo` and then deleting the user from the authentication system.

#### `ActivitiesDBRepositoryImpl.kt`
- **deleteRating**: Removes a user's rating from a specific activity, recalculates the number of ratings and the new mean rating, updates the activity document, and returns a copy of the updated activity.
- **deleteAllUserRatings**: Deletes all ratings given by a specific user across all activities and returns a list of activities with the ratings deleted.

#### `SocialDBRepositoryImpl.kt`
- **deleteAllConversations**: Deletes all conversations that include the specified user.

#### `TokenDBRepositoryImpl.kt`
- **deleteUserToken**: Deletes the specified user's token from the token collection.

#### `UserActivitiesDBRepositoryImpl.kt`
- **deleteUserActivities**: Deletes all activities associated with the specified user.

#### `UserDBRepositoryImpl.kt`
- **deleteUser**: Deletes the specified user's document from the user collection.

#### Related interfaces

I added to the related interfaces these new method signatures. 

## To-Do List

- [x] Create functions related to `activities` collection.
- [x] Create functions related to `conversations` collection.
- [x] Create functions related to `user_activities` collection.
- [x] Create functions related to `user_fcm_tokens` collection.
- [x] Create functions related to `users` collection.
- [x] Test new code on every collection.


app/src/main/java/com/lastaoutdoor/lasta/data/auth/AuthRepositoryImpl.kt
```diff
@@ -92,4 +92,20 @@ constructor(
       emit(Response.Failure(e))
     }
   }
+
+  override suspend fun deleteAccount(): Flow<Response<Boolean>> {
+    return try {
+      val userToDelete = auth.currentUser
```
Andy130604, line 98: You are using a flow as outuput. You should emit loading at the start of the function call.

app/src/test/java/com/lastaoutdoor/lasta/data/auth/AuthRepositoryImplTest.kt
```diff
@@ -251,4 +251,14 @@ class AuthRepositoryImplTest {
       }
     }
   }
+
+  @Test
+  fun `deleteAccount works emits loading then success`() = runTest {
```
Andy130604, line 256: Here you are not testing that it emits loading at all

app/src/test/java/com/lastaoutdoor/lasta/data/db/SocialDBRepoTest.kt
```diff
@@ -377,4 +378,19 @@ class SocialDBRepoTest {
   }
 
    */
+
+  @Test
+  fun `test deleteAllConversations`() = runBlocking {
```
Andy130604, line 383: Use runTest instead of runBlocking

Andy130604 (changes requested):
I liked these function implementations. Particularly the `deleteAllUserRatings` function is really nice. I signaled 2-3 changes that I think would be nice.

app/src/test/java/com/lastaoutdoor/lasta/data/auth/AuthRepositoryImplTest.kt
JeremyHugentobler: +1

JeremyHugentobler (changes requested):
Woaw nice work @Thimphou !! You're coverage is juste incredible for a PR that deals with almost only database calls ! The comments were also really present and of good quality.

I rested some change for the rating being one when no one has given any feedback because it doesn't really feel right. It is debatable so if other people think it is better like this we can just keep it that way.

Thank you for your work

app/src/main/java/com/lastaoutdoor/lasta/data/db/ActivitiesDBRepositoryImpl.kt
```diff
@@ -156,4 +156,64 @@ constructor(context: Context, database: FirebaseFirestore) : ActivitiesDBReposit
...
+      // Recalculate the number of ratings
+      val newNumRatings = activity.numRatings - 1
+
+      // Recalculate the new average rating
+      val newRatings = currentRatings.filter { it["userId"] != userId }
+      val newMeanRating =
+          if (newNumRatings > 0) {
+            newRatings.map { it["rating"].toString().toFloat() }.average()
+          } else {
+            1.0
```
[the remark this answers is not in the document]
Thimphou: Yes I agree with that, but it is implemented like this elsewhere in the code, I just wanted to be consistent with it! What should I do?

app/src/main/java/com/lastaoutdoor/lasta/data/auth/AuthRepositoryImpl.kt
Thimphou: Of course, well done seeing that!

Andy130604 (approved):
Everything seems in order. GJ

app/src/main/java/com/lastaoutdoor/lasta/data/db/ActivitiesDBRepositoryImpl.kt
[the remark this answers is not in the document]
JeremyHugentobler: Yes it was already implemented that way, but still I don't find it very intuitive. 
I think it would be best to just leave it as it is for the moment and discuss it altogether tomorrow and make it a small task for next week if needed.

Thanks for your reply

JeremyHugentobler (approved):
Thanks for the changes !

PR #299, opened 2024-05-22
FIX : suggestion not leaving and Radar error with empty text

JeremyHugentobler:
## Bugs
There was a small bug with the intended behaviour of the app regarding the searchbar and while fixing it, I discovered another one with empty strings.
1. When clicking on a suggestion with the focus not on the textfield (no keyboard shown), the suggestion was applied and then the list cleared. This behaviour was not working if you had the keyboard still opened and clicked on suggestion (the list of suggestions was not cleared)
2. If you wrote something, and then erased it all, it would send an empty string to the API and return an error.

### For the first bug
It was a bit difficult to get my head around it at first, but the problem was due to the focus manager. When clicking a suggestion, I was calling the focusManager to dismiss the keyboard and then rewrite the content of the textfield with the suggestion, but the catch is that clearing the focus manager is not instantaneous ! Therefore the replacement of the text in the texfield was still considered as a user input and caused the list of suggestion to be displayed again...
I fixed it by putting the onclick function of suggestions in a coroutine that has a small delay after clearing the focus and it seems to be working fine now.

### For the second bug
I just added a check in the viewmodel to avoid sending empty strings to the repository and it fixed the issue


Andy130604 (approved):
Small fix so important and yet wasn't yet fixed. GJ keep it up

PR #303, opened 2024-05-23
Offline mode for signup/profile 

npfrei:
still needs testing


Andy130604 (changes requested):
Everything looks good, would still like a bit more comments over the offline DB part. This could improve a bit more its readabiltiy for external code reviewer.

Andy130604 (approved):
Thank you for the comments this will come in handy in the future!

PR #305, opened 2024-05-23
Range Search Screen

JoniDutoit:
## Overview

This PR aims to continue the creation of the radius search feature for finding activities. Actually, since this is quite an important feature for the application, enhancing it will bring the discover screen to the next level! Merging this PR will improve the radius search feature, making it more visually appealing and interactive. 

## Changes

Before | After
---------- | ----------
 |  

- Before: the RangeSearch feature was invoked by a ModalBottomSheet where you could chose your radius without visualization of the selected area,
- After: now, the RangeSearch has become a screen of it's own! Thus, when clicking on the location, now the users are sent to a new interactive screen, composed of a map and a slider to select the desired range between 10-30km .

## Results
https://github.com/LASTA-OUTDOOR/lasta/assets/95919886/7606a3e5-46b4-4967-bae7-d0fb13a906d3

## Testing

> 86% line coverage now. I still have the intention of getting extra line coverage for the RangeSearchScreen if we can allocate time for enhancing line coverage in the future.

## What's next

- Implementing a button to use the current user's location that should be displayed in the RangeSearchScreen by a marker icon could be a nice idea ! 
- Making clusters for activities instead of using activities icons 
- Displaying all activities beforehand instead of only seeing listed activities when selecting the radius


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/components/RangeSearchComposable.kt
```diff
@@ -1,198 +1,202 @@
...
-                          style =
-                              TextStyle(
-                                  fontSize = 22.sp,
-                                  lineHeight = 28.sp,
-                                  fontWeight = FontWeight(400),
-                              ))
-                    }
-              }
-        }
+      Canvas(
```
Andy130604, line 107: I think this is the first time we use this composable. Nice implementation, I like it!

Andy130604 (approved):
This is such a nice feature and was touching so many small files and you still managed to write tests for all the cases. Hats off for your work!

ecornamu (approved):
Well done implementing this upgrade ! Everything looks good to me

PR #306, opened 2024-05-23
Fix: Test UI on different form factors with different languages

JeremyHugentobler:
# UI fix for different form factors
## Method
I used my phone which has an usual format to find as many UI incoherences as possible and did the same with a tablet emulator to stretch the UI and see which parts were not working properly
## Fixes
- Settings screen : 
  - Removed text in front of the selectors. Before there was a label for each selectors that was basically saying the same as what was chosen in the selector, so I removed it and it now looks like those in the filter screen
  - Added some translation
  - Fix the general layout to work with devices with smaller vertical size -> lazycolumn
  - Changed the layout of the button to be displayed nicely when there is not enough horizontal space and removed the changes in font size that were implemented because of the select language (words are usually longer in german)
  - Added a header for the account settings (log out, delete account)
- Filter Screen:
  - Did the same for the vertical space limitation -> lazy column, but I added the fact that the Apply and clear all buttons are always visible (no need to scroll down to apply)


ecornamu (approved):
Well done, code looks good to me :)

PR #308, opened 2024-05-24
Delete account DB functions linked with `Delete Account` button in `SettingsScreen`.

Thimphou:
## Description
This PR implements the Delete Button entirely, using previously coded functions. It links the button to these backend functions. 

## New feature

Now, when you press the `Delete Account` button, everything related to you in DB is now deleted. 
More precisely, it is: 
- Your user page.
- Your friends' friends list is now updated to not contain your userId anymore. 
- Your activities.
- Your messages.
- Your token for notifications.
- Your ratings on activities, and the new computation of ratings is done. 

Also, we reset the user's to default settings. 

I modified the removeAllUsersRating to catch more edge cases about the recalculation of the newRating, and the query was not always the good one (for some reasons, Firestore does not allow a query about a key of a hashmap inside an array, inside a document...), but the idea is quite the same as before.

## Testing

I managed to achieve 84.1% of line coverage.


ecornamu (approved):
Seems all good ! Thank you :)

app/src/main/java/com/lastaoutdoor/lasta/data/auth/AuthRepositoryImpl.kt
```diff
@@ -98,6 +106,11 @@ constructor(
       emit(Response.Loading)
       val user = auth.currentUser
       if (user != null) {
+        userDBRepo.deleteUser(user.uid)
```
Andy130604, line 109: I really like the fact that you put everything in the auth repo, makes it real consise!

Andy130604 (approved):
Great continuation of your work on the DB functions, it was a long-waited feature! This will solve so many problems over data consistency. Great PR, well-tested and well-rounded, I like it!

PR #309, opened 2024-05-25
DiscoverScreen UI Improvement

JoniDutoit:
##Overview

This small PR goal is to enhance the UI design of the discover screen by: 

- Adding an icon and colors to the radius search button, making it more visible & intuitive to users. 
- Improving the Header of the DiscoverScreen so that is looks more clean
- Adding visual aspect to activities's difficulty

Before | After
---------- | ----------
 |


CassioManuguerra (commented):
Nice PR,
Great improvements the UI looks a lot better like this more polished !
I like the new difficulty color,
but now the difficulty at the bottom is redundant.
What do you think about it ?

JoniDutoit:
> Nice PR, Great improvements the UI looks a lot better like this more polished ! I like the new difficulty color, but now the difficulty at the bottom is redundant. What do you think about it ?

I thought of the same thing but I don't feel like it's redudant since new users may not associate the colors to difficulties at first glance. Hence, stating by text the explicit difficulty may help them to better understand all the informations

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/DiscoverScreen.kt
```diff
@@ -356,7 +388,12 @@ fun ActivitiesDisplay(
                       modifier =
                           Modifier.shadow(4.dp, RoundedCornerShape(30))
                               .background(
-                                  MaterialTheme.colorScheme.primary, RoundedCornerShape(10.dp))
+                                  when (a.difficulty) {
```
ecornamu, line 391: You could add this helper function directly in the `Difficulty` class, in case we need this conversion elsewhere

ecornamu (changes requested):
Well done, the UI looks much smoother :) small change request so that you can put the conversion from `Difficulty` to `Color` (Easy to Green, etc...) in a function inside the `Difficulty` class

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/DiscoverScreen.kt
JoniDutoit: It's a great ideia, I accepted this change request! Now, this is an inner function of the enum Difficulty

ecornamu (approved):
Everything is now in order, thanks :)

PR #312, opened 2024-05-25
Set activities' difficulty

CassioManuguerra:
# PR Issue #310
## Overview
Set activities' difficulty by clicking on their difficulty tag in the more info screen, from `EASY` to `HARD`.
Also fixed bug where going on favorite screen, clicking on activity, modifying its difficulty then going back to favorite screen and then clicking again on activity would result in a display bug.
Tested whole new functions (DB, view model, UI)

## What's next
- Add more difficulty levels


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
```diff
@@ -340,15 +341,15 @@ fun Star(iconId: Int) {
...
+fun ElevatedDifficultyDisplay(activityToDisplay: Activity, updateDifficulty: (String) -> Unit) {
   val difficultyColor =
       when (activityToDisplay.difficulty) {
         Difficulty.EASY -> GreenDifficulty
         Difficulty.NORMAL -> YellowDifficulty
         Difficulty.HARD -> RedDifficulty
       }
   ElevatedButton(
-      onClick = { /*TODO let user change activity difficulty */},
+      onClick = { updateDifficulty(activityToDisplay.activityId) },
```
ecornamu, line 352: Well done :) the feature is great !

app/src/main/java/com/lastaoutdoor/lasta/data/db/ActivitiesDBRepositoryImpl.kt
```diff
@@ -101,6 +104,20 @@ constructor(context: Context, database: FirebaseFirestore) : ActivitiesDBReposit
...
 
+  override suspend fun updateDifficulty(activityId: String) {
+    val document = activitiesCollection.document(activityId)
+    val activity = document.get().await()
+    val difficulty = activity.getString("difficulty") ?: "EASY"
+    val newDifficulty =
+        when (difficulty) {
+          "EASY" -> "NORMAL"
+          "NORMAL" -> "HARD"
+          "HARD" -> "EASY"
```
ecornamu, line 115: Maybe you can add this function as a member of the Difficulty class to modularize the code a bit more

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/favorites/FavoritesScreen.kt
```diff
@@ -26,13 +27,20 @@ import com.lastaoutdoor.lasta.ui.screen.discover.ActivitiesDisplay
 fun FavoritesScreen(
     isLoading: Boolean,
     activities: List<Activity>,
+    updateFavorites: () -> Unit,
```
ecornamu, line 30: As a convention I think it would be nice to put all lambda parameters at the end of the parameter list of a function

ecornamu (changes requested):
Many thanks for your work ! 
I request changes so that you put a lambda parameter at the end of the parameter list and modularize a function, but aside from that the feature looks good

app/src/main/java/com/lastaoutdoor/lasta/data/db/ActivitiesDBRepositoryImpl.kt
CassioManuguerra: It's a good idea, however I'm using this only once and it won't be used again so maybe not really needed in this case !

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/favorites/FavoritesScreen.kt
CassioManuguerra: You're right, will do that!

app/src/main/java/com/lastaoutdoor/lasta/data/db/ActivitiesDBRepositoryImpl.kt
```diff
@@ -84,6 +84,9 @@ constructor(context: Context, database: FirebaseFirestore) : ActivitiesDBReposit
     if (!query.isEmpty) {
       val documents = query.documents
       for (document in documents) {
+        if (document.get("startPosition") == null) {
```
JoniDutoit, line 87: I believe that this check isn't necessary and **should** be **removed** as the startPosition would never be null, when it's not valid it is always (0, 0) in the database

JoniDutoit (changes requested):
Great job! The implementation works as expected and it updates all the information correctly, so it's a great addition! I just left one comment that I believe would help to achieve a better code structure. Otherwise, amazing work!

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
CassioManuguerra: Many thanks !

app/src/main/java/com/lastaoutdoor/lasta/data/db/ActivitiesDBRepositoryImpl.kt
CassioManuguerra: you are right, I don't know how it got into my code since I don't remember writing it, but I also found that weird when I saw it, and thought it was someone else who had done this voluntarily. Will remove it now

JoniDutoit (approved):
Nice! Everything looks good to me after the changes

ecornamu (approved):
Good job my lord

PR #313, opened 2024-05-26
Bug fixed the "phantom user" issue

ecornamu:
Prevented the possibility to use the app while logged out by pressing the back button after signing out

It took me some time to tackle out the problem but basically I pop backstack up to and including the main route before getting back to the login screen. 
Thus, the backstack is empty (since the main route is the start of the graph) and the app just closes itself if the user presses the `back` button


Andy130604 (approved):
Nice !

JeremyHugentobler (approved):
It seems to have fixed the issue ! many thanks

PR #319, opened 2024-05-27
Fixes for the MoreInfoScreen

JoniDutoit:
## Overview
Merging this PR will fix many bugs associated to the "View on map" button in the `MoreInfoScreen.kt` from a selected activity. Actually, when activities starting position aren't known from the database, fetching it caused many issues. These ones were fixed thanks to this PR. Furthermore, some issues with the `MoreInfoScreen.kt` itself were also adapted. 

## TODO
- [x] Replaced empty field "duration" from the MoreInfoScreen to the actual distance from the user
- [x] Fixed bug of marker not showing when discovering new position of activity in relation to the database
- [x] Fixed bug of marker being in the ocean when no start position is known
- [x] Fixed bug of map not working correctly from the MoreInfo screen

## Testing
Achieved 82.6% line coverage

## Results 

- FIX: Wrong field called "Duration"

Before | After
----------- | -----------
  |  

- FIX: Marker is not displayed even if good position is found 

Before | After
----------- | -----------
  |  
 
 - FIX: Marker is displayed in the ocean if OSM data starting position is null

Before | After
----------- | -----------
  |  
 
 NOTE: In order to account for missing values in OSM API data, if the starting position is null (= (0.0, 0.0)), then we search a valid position from its itinerary.
 
 - FIX: MoreInfoScreen map doesn't work correctly 

Before | After
----------- | -----------
 <video src="https://github.com/LASTA-OUTDOOR/lasta/assets/95919886/b1c61f42-6444-48ed-9445-7b5d0a9dbf62" width="200"> | <video src="https://github.com/LASTA-OUTDOOR/lasta/assets/95919886/58005952-40d5-4dc7-b1e0-220614d83c37" width="200">


app/src/main/java/com/lastaoutdoor/lasta/viewmodel/MoreInfoScreenViewModel.kt
```diff
@@ -66,11 +69,26 @@ constructor(
...
           val osmData = response.data!!
-          val updatedActivity = activity.copy(startPosition = osmData.getPosition())
+          var newPos = osmData.getPosition()
 
+          if (newPos == Position(0.0, 0.0)) {
+            newPos =
+                when (activity.activityType) {
+                  ActivityType.HIKING,
+                  ActivityType.BIKING -> {
+                    (osmData as Relation)
```
ecornamu, line 79: This fix is really clever ! I get it because you explained it orally but can you comment it so that others can see it as clearly as I do ?

ecornamu (changes requested):
Perfect, this was a much needed fix. Small change request so that you explain your way of avoiding the null position issue through code comments but aside from that it LGTM :)

app/src/main/java/com/lastaoutdoor/lasta/viewmodel/MoreInfoScreenViewModel.kt
JoniDutoit: Yes, thank you for your comment, indeed I should state the changement

JeremyHugentobler (approved):
Thank you for your fixes, it seems to be working much better now, thank you !

ecornamu (approved):
All good <3

PR #324, opened 2024-05-28
Suggest users when adding friends

CassioManuguerra:
# PR Issue #317 
## Overview
This PR introduces the feature to get users suggestions when searching user's email. It implements a new DB and view model function, as well as a new UI interface displaying the suggestions. 

It returns all the users containing the substring, meaning that if you type "ont" it will show you the user "Pierre Dupont" (this choice was made due to there being very little users in our app and the feature was not really useful without this choice). Note that typing "pierre" or "PIErre" will always display "Pierre Dupont".

- The code is 95% tested (except for main nav graph)
- The code is commented
- The code is modularized 

https://github.com/LASTA-OUTDOOR/lasta/assets/91049597/7b8fad6e-3574-46db-9f05-2c4d22ac0327

## What's next
- Implement a different suggestion algorithm
- Add description or mail keyword searching
- Improve UI
- Ask users if they agree to see their mail get shown to anyone


CassioManuguerra:
## Update
In the previous code, users had access to the other users' mail addresses, which was not very confidentiality-friendly.
Updated the code so that now clicking on a suggestion directly sends the friend request, and also made it so that clicking on a suggestion or on the submit button clears the text field.

https://github.com/LASTA-OUTDOOR/lasta/assets/91049597/b6f08304-66e2-4f96-8f0b-861c1772e8b8

ecornamu (approved):
The feature looks so cool, good job !

app/src/main/java/com/lastaoutdoor/lasta/data/db/UserDBRepositoryImpl.kt
```diff
@@ -124,6 +124,53 @@ class UserDBRepositoryImpl @Inject constructor(context: Context, database: Fireb
...
+    val usersToList = mutableListOf<UserModel>()
+    // for each user, check if the username contains the query
+    allUsersUsernames.documents.forEach { doc ->
+      // get the username of the user
+      val userName = doc.getString("userName") ?: ""
+      // if the username contains the query, return the user
+      if (userName.lowercase().contains(query)) {
+        usersToList.add(
+            UserModel(
+                userId = doc.id,
```
JoniDutoit, line 146: This conversion of document firebase to UserModel is nice but it deserves a method of it's own I believe. Check the methods from `UserModel.kt` and `Converters.kt` inside the package models/user to check if there exists already anything useful, and if it doesn't, then add the following method that could be reused then after by someone else!

JoniDutoit (changes requested):
Great work! The UI is looking really cool, and this is a very nice implementation of the feature. I'd just like to point out that maybe you shouldn't be able to see yourself to add as a friend (it doesn't create any error, but it feels weird to me). This is a personal opinion tho, I'd like to know what you think about it! Otherwise, amazing work, I just left one comment about moduralisation of one method, but the rest looks good to me !

app/src/main/java/com/lastaoutdoor/lasta/data/db/UserDBRepositoryImpl.kt
```diff
@@ -124,6 +124,58 @@ class UserDBRepositoryImpl @Inject constructor(context: Context, database: Fireb
...
+    if (allUsersUsernames.isEmpty) {
+      return emptyList()
+    }
+    val usersToList = mutableListOf<UserModel>()
+    // for each user, check if the username contains the query
+    allUsersUsernames.documents.forEach { doc ->
+      // get the username of the user
+      val userName = doc.getString("userName") ?: ""
+      // if the username contains the query, return the user
+      if (userName.lowercase().contains(query)) {
```
JeremyHugentobler, line 148: It's a nice idea to force the usernames to lower case, but therefore it should also be the same for the query variable. Maybe it is done when calling this function, but if we reuse it, it would be more practical and more mistake proof to add a query = query.lowercase()

app/src/main/java/com/lastaoutdoor/lasta/repository/db/UserDBRepository.kt
```diff
@@ -45,6 +45,18 @@ interface UserDBRepository {
    */
   suspend fun addFavorite(userId: String, activityId: String)
 
+  /**
+   * Gets Users corresponding having in their username the given string keyword.
+   *
+   * @param query the string to search for in the username
```
JeremyHugentobler, line 51: you forgot to doc the second and third parameter ^^

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/social/components/AddFriendDialog.kt
```diff
@@ -62,43 +70,80 @@ fun AddFriendDialog(
...
-          KeyboardActions(
-              onDone = {
-                focusManager.clearFocus()
-                requestFriend(text)
-              }))
+  Column(modifier = Modifier.fillMaxWidth(), horizontalAlignment = Alignment.CenterHorizontally) {
+    TextField(
+        value = text,
+        onValueChange = { newValue -> text = newValue },
+        label = { Text(LocalContext.current.getString(R.string.email)) },
```
JeremyHugentobler, line 97: If we now search by username, it doesn't make a lot of sense to write email in the textfield

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/social/components/AddFriendDialog.kt
```diff
@@ -62,43 +70,80 @@ fun AddFriendDialog(
...
+                  requestFriend(text)
+                }))
+
+    // Fetch friend suggestions when text changes
+    LaunchedEffect(text) {
+      // put text in small letters
+      fetchFriendSuggestions(text.lowercase())
+    }
+    // Display friend suggestions
+    LazyColumn(
```
JeremyHugentobler, line 113: This is a very long composable, maybe you could extract a different one for the list to make it lighter ro read ?

JeremyHugentobler (changes requested):
Hi Cassioo !

Nice work, the result look incredibly good ! There are some small changes that I requested and you can find them with the different comments I left in the review, but overall it's an amazing feature thank you !!

app/src/main/java/com/lastaoutdoor/lasta/data/db/UserDBRepositoryImpl.kt
CassioManuguerra: You're right, for other calls it's a good thing to do

app/src/main/java/com/lastaoutdoor/lasta/repository/db/UserDBRepository.kt
CassioManuguerra: yes thanks for noting I just updated it

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/social/components/AddFriendDialog.kt
CassioManuguerra: I mean it still works with email addresses and it's a good thing that it does because the check is done via email, so I think it is relevant to leave this there

JeremyHugentobler: Yes but how can the user know that it is possible to just enter names ? (of course he can try, but the goal of an indication in a textfield is to tell what is expected in the field) 

Would it be okay to write Email OR user name ?

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/social/components/AddFriendDialog.kt
CassioManuguerra: uWu it really does not seem really long to me, compared to others particularly

app/src/main/java/com/lastaoutdoor/lasta/data/db/UserDBRepositoryImpl.kt
CassioManuguerra: Yep changed it!

CassioManuguerra:
# Update
Last update, removed the submit button, so that there is no email adding, you add friends by their account, the friend adding feels more pro now

https://github.com/LASTA-OUTDOOR/lasta/assets/91049597/68bfe749-ab44-4beb-b7c1-bd02c1479d0d

JeremyHugentobler (approved):
Thanks for your changes !

PR #325, opened 2024-05-28
UI fixes concerning when there is no "something". 

### LINKED ISSUE #322: Redesign discover screen to explain when no activities are displayed / no friends request are displayed

Thimphou:
## Objective
This PR aims to improve the app's design when: 
- There is no activities found. 
- The user has no friend requests. 

### DiscoverScreen

Before, there was nohting showed on screen when no activities were fetched, now the user knows what is going on as you can see below! 

### FriendsRequestList

Improved UI of this page. 

| Before                                      | After                                       |
|---------------------------------------------|---------------------------------------------|
|  |  |

### MoreInfoScreen

Quick fix concerning the basic icons of activities (previously it was markers, now it's only what's inside the markers), but it is very straightforward.


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/DiscoverScreen.kt
```diff
@@ -115,7 +119,7 @@ fun DiscoverScreen(
           if (discoverScreenState.isLoading) {
             LoadingAnim(width = 35, tag = "LoadingBarDiscover")
           } else if (discoverScreenState.activities.isEmpty()) {
-            /* TODO */
+            EmptyActivityList()
```
JoniDutoit, line 122: Nice! This Todo was annoying me hahah

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/discover/DiscoverScreen.kt
```diff
@@ -445,3 +449,22 @@ fun ActivitiesDisplay(
...
+      horizontalAlignment = Alignment.CenterHorizontally) {
+        Icon(
+            painterResource(id = R.drawable.not_found),
+            contentDescription = "No Activities",
+            modifier = Modifier.size(85.dp).testTag("NoActivitiesLogo"))
+        Text(
+            text = LocalContext.current.getString(R.string.no_activities),
+            style =
+                TextStyle(
+                    fontWeight = FontWeight.Normal, fontSize = 20.sp, textAlign = TextAlign.Center),
```
JoniDutoit, line 467: For the text, is better to use MaterialTheme.typography package for different sort of texts so that it looks homegenous through out the app! Hardcoded text fonts wont adapted themselves with the screen size which is an UI issue

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/moreinfo/MoreInfoScreen.kt
```diff
@@ -456,7 +462,7 @@ fun ActivityPicture(activityToDisplay: Activity) {
       Image(
           painter = painterResource(id = defaultId),
           contentDescription = "Default Activity Picture",
-          modifier = Modifier.padding(5.dp).size(90.dp))
+          modifier = Modifier.padding(5.dp).size(90.dp).testTag(testTag))
```
JoniDutoit, line 465: The activity icon looks rather big on my screen, did you try with smaller sizes to see how it goes ?

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/social/components/FriendsRequestList.kt
```diff
@@ -41,18 +47,38 @@ fun FriendsRequestList(
...
+          verticalArrangement = Arrangement.Center,
+          horizontalAlignment = Alignment.CenterHorizontally) {
+            Icon(
+                painterResource(id = R.drawable.no_friends),
+                contentDescription = "No Activities",
+                modifier = Modifier.size(85.dp).testTag("NoFriendsLogo"))
+            Text(
+                text = LocalContext.current.getString(R.string.no_friends_yet),
+                style =
+                    TextStyle(
```
JoniDutoit, line 76: Here again the font is unfortunately hard coded with size 20.sp and wont adapted to the screen size, consider changing it to a font from the package MaterialTheme.typography provided by Android

JoniDutoit (changes requested):
Amazing work! Really like the icons you added, also good idea changing the icon from the MoreInfoScreen to the rounded one instead! It does indeed add information for a better UX experience. I left some suggestions so that you could improve some aspects of your implementation, otherwise it looks good to me!

Thimphou:
Thanks @JoniDutoit for these useful remarks! I changed the size of the icons on the MoreInfoScreen, adapted the TextStyle previously used into MaterialTheme, and I also added an Icon on Feed and Friends tabs on Social Screen, and modularized it!

Hope these changes will make you happy :)

JoniDutoit (approved):

PR #328, opened 2024-05-29
Fixed setup screen size issue

JeremyHugentobler:
## Bug description
The menu was not in a lazycollumn and values were hardcoded, which caused the finish button to be hidden on certain devices (my phone hihi). 

## Bug resolution
I created a LazyList with the setup menu iniside, which makes it scrollable and fix the issue.


Thimphou (approved):
Nice changes to make the UI of this page more aesthetic ;) Thanks Jeremy!

PR #329, opened 2024-05-29
Did basic UI for tracking as well as implemented the sensors and stopwatch with it.

Andy130604:
Still testing


app/src/main/java/com/lastaoutdoor/lasta/utils/Constants.kt
```diff
@@ -0,0 +1,18 @@
...
+  const val STOPWATCH_STATE = "STOPWATCH_STATE"
+
+  const val NOTIFICATION_CHANNEL_ID = "STOPWATCH_NOTIFICATION_ID"
+  const val NOTIFICATION_CHANNEL_NAME = "STOPWATCH_NOTIFICATION"
+  const val NOTIFICATION_ID = 10
+
+  const val CLICK_REQUEST_CODE = 100
+  const val CANCEL_REQUEST_CODE = 101
+  const val STOP_REQUEST_CODE = 102
+  const val RESUME_REQUEST_CODE = 103
```
ecornamu, line 17: Well done ! A `Constants` class is ideal in this heavyly-abstract context

ecornamu (approved):
Well done ! Its a shame that you can't test the whole but we cannot afford to put time into testing this close to the deadline so I'll approve anyway :)

app/src/main/java/com/lastaoutdoor/lasta/utils/Constants.kt
Andy130604: Thank you, thought it would be useful in this case

PR #330, opened 2024-05-29
Fixes on the social screens

JeremyHugentobler:
## Social screen fixes

Those are the bugs I found:
- Weird behaviour with already started conversation that were not displayed in the list. 
- Existing conversation, but still empty were shown in the list (not really useful)
- The share in-app functionnality was not working if you didn't have a previous conversation with the friend

I fixed all of them and I added the fact that when sharing in app, you are redirected to the conversation -> it gives a better feedback that it has been sent


app/src/main/java/com/lastaoutdoor/lasta/ui/screen/social/components/MessageList.kt
```diff
@@ -66,8 +67,17 @@ fun MessageList(
...
-        items(messages.size) {
-          MessageCard(messages[it], friends[it].userId, navigateToConversation)
+        items(messages.size) { messageIndex ->
+          val friendId =
+              messages[messageIndex]
+                  .members
+                  .firstOrNull { friends.map { it.userId }.contains(it.userId) }
+                  ?.userId
+          if (friendId == null) {
+            Log.e("MessageList", "Friend not found in friends list")
```
ecornamu, line 77: Don't forget to remove the debug console prints

ecornamu (approved):
Good job !

app/src/main/java/com/lastaoutdoor/lasta/ui/screen/social/components/MessageList.kt
JeremyHugentobler: It's not really a print, I just wanted to log an error (you can see it as non-blocking throw)

Andy130604 (approved):
Everything looks good to me !

PR #334, opened 2024-05-29
Offline mode fixes

npfrei:
Added some fixes to the Offline mode : 
-Fixed some issues with error toasts
-Buttons that are not usable in offline are now disabled are grayed out
-Others small fixes


Andy130604 (approved):
This is a hard feature to see in the app, but you implemented it really neatly!

JeremyHugentobler (approved):
thanks for your work

PR #335, opened 2024-05-29
Swippable tabs for the favorite screens

JeremyHugentobler:
## Swippable tabs
It is now possible to do a swipe gesture tu navigate through the different pages of the social tabs.

### Issues encountered
I encountered some issues with the icons blinking due to animation delays and scroll times


ecornamu (approved):
Carré la zone

PR #338, opened 2024-06-02
Working share function out of the app

JeremyHugentobler:
### Content
When clicking on the share button, you can now send a link to a friend, which is recognized by the application and can be opened with ! (Warning, there is no website, so if you click on the link and don't have the app, you will just have a 404 on your browser).
You need to be logged in to be redirected to the more info screen of the activity.

### Emulator troubleshooting
The link is officially linked to the application, but with "the release" key of the project. If you try to click on a share link on your emulator, which is linked to **your debug key**, it won't be recognized and therefore will be launched in the browser. To fix that, you can long press the app icon, go to it's details and allow the app to open link of the form "lasta.jerem.ch".

Another solution is to use the APK I generated with the release key. You can find it here : https://github.com/LASTA-OUTDOOR/lasta/actions/runs/9340089704 (With this APK, link are automatically associated to the app since it uses the right signature/key)

### Testing
Most of the work was in the navgraph and outside of the app, and is therfore hardly testable. 

### Final note
It took me a lot of time to have this feature working, even though there is only a few lines of code, I had a to try a loooot of things to beat the issues I encountered. (NavigateBack button that has to know if you arrived on the page using a link or directly by navigating in the app, block the process if you are not logged in, transfer arguments in multiple navgraphs, ...)


npfrei (changes requested):
Hello, your testing is under 80%, do you think there is a way to increase it ?

JeremyHugentobler:
> Hello, your testing is under 80%, do you think there is a way to increase it ?

Hi, As I wrote in the PR description, this new feature doesn't really add some composable or backend code to test. It is just some logicial conditions and actions in the navgraph and some config which is barely testable and I can therefore not achieve the 80%. It has been the same for all previous PR on navgraphs

npfrei (approved):
Okay, we already had issues with testing the navgraph previously, so i will accept it

PR #342, opened 2024-06-02
Fixed a bug concerning the Delete account. 

### LINKED ISSUE #341: General debugging of the app

Thimphou:
Hey!

This PR solves a bug, concerning a rare bug of Delete Account, occuring only when there is a JobCancellationException, which was happening sometimes (concurrency...). I solved it by putting everything in async blocks, and await all jobs at the end. Also, I removed the SignOut call in the deleteAccount.


npfrei (approved):
Great PR to fix a really problematic bug, good Job !

